PR #3, opened 2018-10-22
migrations: biting ids and new ext_ref column

bzwei:
Change all table IDs to `bigint`.
Add ext_ref column to `templates`. This is the linkage between BPM and approval service.


bzwei:
@ManageIQ/bots assign @gmcculloug 
cc @hsong-rh

db/migrate/20181022183022_change_ids_to_bigint.rb
```diff
@@ -0,0 +1,8 @@
+
+class ChangeIdsToBigint < ActiveRecord::Migration[5.1]
+  def change
+    %i{actions groups requests stages templates workflowgroups workflows}.each do |table|
+      change_column table, :id, :bigint
+    end
+  end
```
gmcculloug: @bzwei I think you and @hsong-rh are the only ones with this database at the moment.  If you guys can drop then recreate the database the only thing you need to do to get bigint id columns is to remove `id: serial` from `db/migrate/0_init_tables.rb`.

I would also recommend, in a different PR, updating gem versions for Rails and pg.  See https://github.com/ManageIQ/service_portal-api/pull/22

bzwei:
@gmcculloug done

PR #5, opened 2018-10-23
Seed default template

bzwei:
Use seeding method to create a default template called `Basic`. The `ext_ref` attribute is subject to future update. We may allow the BPM to update this attribute during its startup procedure.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -1,7 +1,5 @@
-# This file should contain all the record creation needed to seed the database with its default values.
-# The data can then be loaded with the rails db:seed command (or created alongside the database with db:setup).
-#
-# Examples:
-#
-#   movies = Movie.create([{ name: 'Star Wars' }, { name: 'Lord of the Rings' }])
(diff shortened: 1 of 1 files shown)


bzwei:
@gmcculloug @hsong-rh

db/seeds.rb
```diff
@@ -1,7 +1,5 @@
-# This file should contain all the record creation needed to seed the database with its default values.
-# The data can then be loaded with the rails db:seed command (or created alongside the database with db:setup).
-#
-# Examples:
-#
-#   movies = Movie.create([{ name: 'Star Wars' }, { name: 'Lord of the Rings' }])
-#   Character.create(name: 'Luke', movie: movies.first)
+Template.create(
+  :title => 'Basic',
+  :description => 'The very basic approval workflow that supports multi-level approver groups through email notification',
```
gmcculloug: Suggest changing the start to "A basic approval ...".

If it was "very basic" it would not include multi-level approver groups.  😄

PR #6, opened 2018-10-24
Fix has_many :groups

bzwei:
When doing a simple query `workflow.groups` an error occurs:
```
ActiveRecord::HasManyThroughOrderError: Cannot have a has_many :through association 'Workflow#groups' which goes through 'Workflow#workflowgroups' before the through association is defined.
```
It turns out `has_many :workflowgroups` must appear before `has_many :groups`.


bzwei:
@miq-bot assign @gmcculloug 
cc @hsong-rh

PR #7, opened 2018-10-25
Send events when stage or request state changes

bzwei:

PR #8, opened 2018-10-28
Auto approve requests

bzwei:
By default any request through POST will remain in pending state.
With environment variable `AUTO_APPROVAL` set to  `y`, the request will be auto approved.
Environment variable  `AUTO_APPROVAL_INTERVAL` will be used to insert delays to each approver level to simulate manual approver process, for example
```
export AUTO_APPROVAL_INTERVAL=10
```
inserts 10 seconds delay for each approver from state being `notified` to `finished`, and the result will be `approved`.

Depends on #7


mkanoor:
@bzwei @gmcculloug 
Isn't there supposed to be an Approval workflow called Auto-Approve that the UI could use if it wants to use Auto Approval instead of using the ENV

gmcculloug:
Discussed with @bzwei  and @mkanoor and this use of the environment variable is for testing-only at this stage while we continue to develop the service.

Also, we'll likely need to rename the service classes here in a future refactoring once we decide on the naming and namespacing of those classes.  But I think it will likely follow something like what is described here: http://brewhouse.io/blog/2014/04/30/gourmet-service-objects.html

PR #9, opened 2018-11-02
Add relations into swagger file

hsong-rh:
This PR contains following changes:
1. Add id property to each model definition in swagger yaml;
2. Add relationship between workflow and group in swagger yaml;
3. Add missed description fields in swagger yaml;
4. Add mixins for user and approver;


hsong-rh:

app/controllers/approvers_controller.rb
```diff
@@ -8,47 +8,6 @@
 Generated by: https://github.com/swagger-api/swagger-codegen.git
 
 =end
-class ApproversController < ApplicationController
+class ApproversController < AdminsController
```
bzwei: Approver has more privilege than admin?

public/doc/swagger-2.yaml
```diff
@@ -843,12 +865,18 @@ definitions:
...
-      - template
     properties:
+      id:
+        type: string
+        readOnly: true
       name:
         type: string
       description:
         type: string
+      groups:
```
bzwei: You can add `group_ids` which is an array of ids. This is helpful to create a new workflow.

app/controllers/base_controller.rb
```diff
@@ -0,0 +1,34 @@
...
+OpenAPI spec version: 1.0.0
+
+Generated by: https://github.com/swagger-api/swagger-codegen.git
+
+=end
+class BaseController < ApplicationController
+
+  def add_request
+    req = RequestCreateService.new(params[:workflow_id]).create(request_params)
+    json_response(req, :created)
```
mkanoor: @hsong-rh 
The swagger-2.yaml is missing the request object definition coming back from addRequest

mkanoor: @hsong-rh 
https://github.com/ManageIQ/approval-api/blob/408db9301c937bc2065b3e94da960721be05c39c/public/doc/swagger-2.yaml#L308

We need a 200 and response object back

app/controllers/approvers_controller.rb
bzwei: but `ApproversController` is a subclass of `AminsController`, which seems opposite to what you claimed.

You may extract common code into a mixin to avoid this confusion.

hsong-rh: I created mixins for user and approver.

app/controllers/base_controller.rb
hsong-rh: Response object is added with 201, the normal return code for successful `created` operation.

mkanoor:
👍 Tested adding a request it returns a Request object

spec/requests/requests_spec.rb
```diff
@@ -111,13 +111,13 @@
...
 
-  # Test suite for PUT /workflows/:workflow_id/requests
+  # Test suite for POST /workflows/:workflow_id/requests
   describe 'POST /workflows/:workflow_id/requests' do
     let(:item) { { 'disk' => '100GB' } }
-    let(:valid_attributes) { { requester: '1234', name: 'Visit Narnia', content: JSON.generate(item), decision: 'unknown', state: 'pending' } }
+    let(:valid_attributes) { { requester: '1234', name: 'Visit Narnia', content: JSON.generate(item) } }
 
     context 'when request attributes are valid' do
-      before { post "/workflows/#{workflow_id}/requests", params: valid_attributes, headers: admin_encode_key }
```
syncrou: Was the header constraint pulled out in an earlier PR?

hsong-rh: Post a request is user's operation. It shouldn't have admin header constraint. I corrected it here.

PR #10, opened 2018-11-13
Add missing information for approval APIs client

hsong-rh:
Changes to enable approval client to retrieve more detailed informations.


hsong-rh:
@bzwei @syncrou @mkanoor Please review.

bzwei:
I don't understand why there are schema and migration changes. There should be only swagger spec modeling changes.

hsong-rh:
@bzwei Have to change the schema, otherwise auto generated client will always display nil for associations. Will do research to see if other ways can do this.

mkanoor:
@hsong-rh 
How does the client navigate the associations?
Should it just be a API path
/approval/{approval_id}/messages?

hsong-rh:
@mkanoor Should be like approval_instance.messages, ex. request.stages

hsong-rh:
@bzwei Changed to untouch schema file.

mkanoor:
@hsong-rh 
Would the caller have to always fetch the whole approval_request object or would it be able to get to the attributes which are an array separately

/approval/{approval_request_id}/messages

hsong-rh:
@mkanoor 
No, it depends on what you want to fetch. This PR is only enabling the association between models in client side. For example:
```
=> #<ApprovalAPIClient::Workflow:0x007ffc55016518
 @group_ids=["1", "2", "3"],
 @groups=
  [#<ApprovalAPIClient::Group:0x007ffc540ac0a0 @contact_method="email", @contact_setting="group1@example.com", @id="1", @name="group_1">,
   #<ApprovalAPIClient::Group:0x007ffc5287e960 @contact_method="email", @contact_setting="group2@example.com", @id="2", @name="group_2">,
   #<ApprovalAPIClient::Group:0x007ffc550967b8 @contact_method="email", @contact_setting="group3@example.com", @id="3", @name="group_3">],
 @id="1",
 @name="workflow_1">
```
In client side, after we fetched workflow instance, its association with groups/group_ids is auto-populated too.

app/models/request.rb
```diff
@@ -27,4 +27,8 @@ class Request < ApplicationRecord
   scope :decision, ->(decision) { where decision: decision }
   scope :state, ->(state) { where state: state }
   scope :requester, ->(requester) { where requester: requester }
+
+  def as_json(options={})
+    super(:include => [ :workflow, :stages ])
```
bzwei: In general we don't need to expose `belong_to` relation. Common practice is to provide a `links` sections in the payload; else we can simply include the `workflow_id`. 
Because stages are so tightly integrated with request we can treat stages as embedded although they are stored in another table, we can do `super(:include => [:stages])`

hsong-rh: It's no way to provide such a link through swagger spec. I will simply include workflow_id in response. Similar logic will apply to other models: just include the associated ids.

hsong-rh:
@bzwei Please merge this PR.

hsong-rh:
@Fryguy I don't know how to fix above Hakiri errors. It's said all these warnings are coming from dependent gems. Can you take a look? thanks.

PR #11, opened 2018-11-14
Build a dev docker image

bzwei:
To build a docker image:
```
docker build -t service_catalog --file=DevDockerfile .
```

To run the docker container:
```
docker run -p 5000:5000 --name=approval_service -e DATABASE_URL="postgres://myuser:mypass@0.0.0.0/service_approval_development" -d approval_service
```
In the above command line you need to feed in the correct setting to connect to a database for approval service. Database username and password are optional if your testing Postgresql does not require them.


bzwei:
@syncrou @gmcculloug @hsong-rh

config/database.yml
```diff
@@ -6,7 +6,7 @@ default: &default
   password: <%= ENV['SERVICE_APPROVAL_DATABASE_PASSWORD'] %>
   wait_timeout: 5
   min_messages: warning
-  host: localhost
+  host: <%= ENV['SERVICE_APPROVAL_DATABASE_HOST'] %>
```
syncrou: `database.yml` should be replaced by a sample [database.dev.yml](https://github.com/ManageIQ/service_portal-api/blob/master/config/database.dev.yml) so the real database.yml doesn't create dirty commits.

syncrou: You can then replace the `<%= ENV['SERVICE_APPROVAL_DATABASE_HOST'] %>` environment variable in the `docker run` command with `DATABASE_URL`.

bzwei: my goal is to have a `database.yml` that does not need modification.

config/ssl/localhost.key
```diff
@@ -0,0 +1,28 @@
+-----BEGIN PRIVATE KEY-----
```
carbonin: Don't commit private keys.

DevDockerfile
```diff
@@ -1,21 +1,33 @@
...
-
-WORKDIR $RAILS_ROOT
-
-COPY Gemfile Gemfile
-
-RUN gem install bundler
-
-RUN bundle install
-
+FROM manageiq/ruby:latest
```
carbonin: This looks almost exactly like `Dockerfile` in this repo. If the image is going to be nearly identical, why have two?

config/ssl/localhost.key
carbonin: If you need to generate a cert at build or runtime [here](https://github.com/ManageIQ/manageiq-appliance/blob/master/LINK/usr/bin/generate_miq_server_cert.sh) is an example of how we do it for manageiq

bzwei:
no longer needed

PR #12, opened 2018-11-19
Refactor event service to use constants and env

bzwei:
Allow to set Kafka topic through `ENV['KAFKA_TOPIC']`

Some refactoring to use constants.

### DIFF
--- a/app/services/event_service.rb
+++ b/app/services/event_service.rb
@@ -1,7 +1,11 @@
 require 'manageiq-messaging'
 
 class EventService
-  APPROVAL_EVENT_TOPIC = 'insights-approval-service'.freeze
+  EVENT_REQUEST_STARTED  = 'request_started'.freeze
+  EVENT_REQUEST_FINISHED = 'request_finished'.freeze
+  EVENT_APPROVER_GROUP_NOTIFIED = 'approver_group_notified'.freeze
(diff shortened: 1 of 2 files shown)


hsong-rh:
LGTM 👍

Fryguy:
bounced the PR to allow Travis and CodeClimate to run now that it's ben merged.

app/services/event_service.rb
```diff
@@ -1,7 +1,11 @@
 require 'manageiq-messaging'
 
 class EventService
-  APPROVAL_EVENT_TOPIC = 'insights-approval-service'.freeze
+  EVENT_REQUEST_STARTED  = 'request_started'.freeze
```
mkanoor: @bzwei 
Any reason for using snake case for event names makes it look like a variable. 
Could we just use camel case
**RequestStarted** instead or 
use all upper case with under scores.

**REQUEST_STARTED** which makes it feel like constants from the windows messaging world.

Also can this be stored some place common for other micro services to use.

bzwei: Did a search on manageiq codebase. All constants are snake case all capitals. 

It is good to store constants in a common place, but we don't have such a place right now.

mkanoor:
👍

PR #20, opened 2018-11-20
Remove Gemfile.dev.rb from bundler.d directory

carbonin:
This was causing an issue with the docker build where pry would
not be installed at first because we bundle with only the Gemfile
present, but then when the required gems are determined at runtime
bundler fails because pry isn't present


Fryguy:
Oops 😅 

Can you also add the bundler.d directory to the .gitignore?

carbonin:
Pretty sure this didn't break that workflow spec ... :thought_balloon:

Fryguy:
Let's see if it's a sporadic failure

PR #21, opened 2018-11-21
Add version support

hsong-rh:
Add API version support.

https://github.com/ManageIQ/approval-api/issues/14


hsong-rh:
@bzwei @syncrou @mkanoor Please review.

app/controllers/api/v0/admins_controller.rb
```diff
@@ -0,0 +1,167 @@
...
+      end
+
+      def fetch_workflow_by_id
+        workflow = Workflow.find(params[:id])
+
+        json_response(workflow)
+      end
+
+      def fetch_workflow_requests
+        workflow = Workflow.find(params[:workflow_id])
```
mkanoor: @hsong-rh 
Maybe in a different PR we should use params.require(:workflow_id) this is part of the strong parameters which will ensure that the parameter exists and will raise an error if the parameter is missing.  **ActionController::ParameterMissing** will be raised 
http://signalexception.com/blog/2013/07/14/customizing-the-strong-parameters-missing-parameter-response.html
Rails converts this to a 400 HTTP error code

hsong-rh: @mkanoor Ok, I'll open another PR to use params.require for these find operations.

PR #23, opened 2018-11-28
Changed to used params.require to get attributes

hsong-rh:
Use params.require to ensure the parameter exists, otherwise will raise an **ActionController::ParameterMissing** error.


hsong-rh:
@bzwei @mkanoor @syncrou Please review.

PR #27, opened 2018-11-30
Create an auto-approve workflow that does not need a template

bzwei:
This auto approve workflow will always approve a request without sending notification to any human approvers. It stills sends Kafka events so that the portal service can rely on approval events to drive the catalog ordering.

The workflow will be available out of box; therefore a seeding is required. It does not need a template because it is one-of-a-kind.


spec/services/request_create_service_spec.rb
```diff
@@ -16,15 +16,15 @@
     end
   end
 
-  describe '#create' do
+  context 'simulate human approvals' do
```
hsong-rh: To my understanding, this and next context are both auto approval test cases, one is with normally created workflow, the other uses seed-generated workflow. I feel `human approval` is bit of misleading here.

hsong-rh:
LGTM 👍

app/services/request_create_service.rb
```diff
@@ -21,15 +21,23 @@ def create(options)
...
+  end
+
+  def auto_approve?
+    ENV['AUTO_APPROVAL'] && ENV['AUTO_APPROVAL'] != 'n'
+  end
+
   def start_approval_process(request)
     Thread.new do
-      auto_approve(request)
+      default_approve? ? default_approve(request) : auto_approve(request)
```
hsong-rh: Question about `default_approve`: what's its purpose? Seems like `auto_approve` is more similar to the real case.

bzwei: I changed the seeded workflow from `auto_approve` to `always_approve` and use it as a default. As indicated by the name this workflow always approves unconditionally; no approver groups are needed.

The concept of "auto approve" is reserved for the use of algorithm approval based on business rules. The result could be approved or denied. Currently `auto_approve` controlled by environmental variables is to simulate multi-level approvals. It will not be used in production.

hsong-rh: 👍

PR #28, opened 2018-11-30
Replaced PUT with PATCH action

hsong-rh:
Based on the legacy definition, PUT is used to update all attributes of the resource, and PATCH is for individuals. In this case Approval API should use PATCH, rather than PUT, to update resources.


config/routes.rb
```diff
@@ -10,7 +10,7 @@
 =end
 Rails.application.routes.draw do
 
-  mount SwaggerUiEngine::Engine, :at => '/open-api'
+  mount SwaggerUiEngine::Engine, at: '/api'
```
gmcculloug: This seems unrelated to the Put/Patch change and has switched back and forth a few times on this file.  service_portal-api is currently using `open-api` https://github.com/ManageIQ/service_portal-api/blob/master/config/routes.rb#L13

What's the correct value here?

cc @syncrou

syncrou: I changed it to `open-api` because of the versions pr.  The base path `/api/*` wasn't being recognized correctly by the rails engine, plus I wanted to differentiate between api calls `/adi` and Swagger UI `/open-api`

config/routes.rb
```diff
@@ -10,7 +10,7 @@
 =end
 Rails.application.routes.draw do
 
-  mount SwaggerUiEngine::Engine, :at => '/open-api'
+  mount SwaggerUiEngine::Engine, at: '/open-api'
```
bzwei: why not rollback to `:at => '/open-api'` so that this diff does not appear at all?

bzwei:
@hsong-rh we may need another PR to clean up all endpoints. Some may not support full CRUD.

config/routes.rb
hsong-rh: Done.

hsong-rh:
@bzwei Yes, I'll create a PR to cleanup some extra endpoints.

gmcculloug:
@hsong-rh @bzwei Please add a more detail into the description as to what promoted this change.  It helps reviewers and future readers to understand the reasoning.  Thanks.

hsong-rh:
@gmcculloug The description is updated with more details.

PR #30, opened 2018-12-06
Add ActionCreateService

bzwei:
Create a service class for adding a new action.
* A request has many stages (multi-level approval). Stage entities are created by backend and read-only through API. 
* A request approval status is determined by the statuses of stages. It is read-only through API.
* Action is the only type can be created by approver. Once created it is read-only.
* Depends on the operation type of an action, the backend automatically updates the status of stage and request.

This PR reinforces the above policy.


app/services/stage_update_service.rb
```diff
@@ -9,10 +9,50 @@ def update(options)
...
     return if old_state == stage.state
-    if stage.state == Stage::NOTIFIED_STATE
+    case stage.state
+    when Stage::NOTIFIED_STATE
       EventService.new(stage.request).approver_group_notified(stage)
-    elsif stage.state == Stage::FINISHED_STATE
+      request_started if first_stage?
+    when Stage::FINISHED_STATE
       EventService.new(stage.request).approver_group_finished(stage)
+      request_finished(stage.decision, stage.reason) if last_stage?
```
hsong-rh: Stage may have either `approved` or `denied` status when its state is `finished`. If it's `denied`, `request_finished` should also be called even it's not the last stage, right?

app/services/stage_update_service.rb
```diff
@@ -9,10 +9,50 @@ def update(options)
...
+    when Stage::SKIPPED_STATE
+      last_stage_skipped if last_stage?
```
hsong-rh: If stage's status is marked as `denied`, all following stages should mark their states to `SKIPPED_STATE`, or just break and update request's state directly. I didn't see this logic.

How can stage have `Stage::SKIPPED_STATE`?

app/services/stage_update_service.rb
bzwei: We don't want to complete the request immediately because all stages are already created. We need to skip all remaining stages before we can complete the request. 
It is BPM's responsibility to set all remaining stage's state to skipped.

app/services/stage_update_service.rb
bzwei: See my comments above. BPM should set stage to skipped.

gmcculloug:
@hsong-rh Please review latest changes.

app/services/stage_update_service.rb
hsong-rh: 👍

app/services/stage_update_service.rb
hsong-rh: In this case BPM will update stage via rest api client, we should keep those stage PATCH endpoints.

hsong-rh: Now I understand. BPM will create an action to update its stage's state.  Above comment can be ignored.

PR #33, opened 2018-12-12
Hardcode the basePath in the swagger yml for client generation

Fryguy:


Fryguy:
@carbonin Please review.

PR #34, opened 2018-12-13
Change endpoint to create workflow with groups

hsong-rh:
Change the POST action of workflow. The input parameters contain an array of approval group ids.

### DIFF
--- a/app/controllers/api/v0/admins_controller.rb
+++ b/app/controllers/api/v0/admins_controller.rb
@@ -18,8 +18,10 @@ module Api
       end
 
       def add_workflow
-        workflow = Template.find(params.require(:template_id)).workflows.create!(workflow_params)
+        workflow = WorkflowCreateService.new(params.require(:template_id)).create(workflow_params)
         json_response(workflow, :created)
+      rescue ActiveRecord::RecordNotFound => e
+        json_response({ :message => e.message }, :unprocessable_entity)
       end
 
       def fetch_group_by_id
@@ -116,11 +118,7 @@ module Api
       end
 
       def workflow_params
-        params.permit(:name, :description)
-      end
-
-      def request_params
-        params.permit(:name, :requester, :content)
+        params.permit(:name, :description, :group_ids => [])
       end
     end
   end
--- a/app/services/workflow_create_service.rb
+++ b/app/services/workflow_create_service.rb
@@ -1,0 +1,11 @@
+class WorkflowCreateService
+  attr_accessor :template
+
+  def initialize(template_id)
+    self.template = Template.find(template_id)
+  end
+
+  def create(options)
+    template.workflows.create!(options)
+  end
+end
--- a/public/doc/swagger-2.yaml
+++ b/public/doc/swagger-2.yaml
@@ -603,7 +603,7 @@ paths:
           description: Parameters need to create workflow
           required: true
           schema:
-            $ref: '#/definitions/Workflow'
+            $ref: '#/definitions/WorkflowParameters'
       responses:
         '201':
           description: Success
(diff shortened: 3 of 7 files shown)


hsong-rh:
@bzwei Please review.

public/doc/swagger-2.yaml
```diff
@@ -697,8 +697,16 @@ securityDefinitions:
   approval_auth:
     type: basic
     description: The user id/password is needed when accessing this API externally
```
bzwei: do we support user/password type authentication? if not, let's remove it for now.

app/controllers/api/v0/admins_controller.rb
```diff
@@ -18,8 +18,14 @@ def add_group
...
 
       def add_workflow
+        group_ids = params.require(:group_ids)
+        groups = Group.find(group_ids)
+
         workflow = Template.find(params.require(:template_id)).workflows.create!(workflow_params)
+        workflow.groups << groups
         json_response(workflow, :created)
+      rescue ActiveRecord::RecordNotFound => e
+        json_response({ :message => e.message }, :unprocessable_entity)
```
bzwei: with your enhancement in `exception_handler` do you still need this rescue?

app/controllers/api/v0/admins_controller.rb
```diff
@@ -18,8 +18,14 @@ def add_group
       end
 
       def add_workflow
+        group_ids = params.require(:group_ids)
```
bzwei: Move this code to a service class `WorkflowCreateService`

public/doc/swagger-2.yaml
```diff
@@ -861,6 +869,22 @@ definitions:
         type: string
         description: Associated template id
         readOnly: true
+  WorkflowGroups:
```
bzwei: consider name `WorkflowParameters`?

public/doc/swagger-2.yaml
```diff
@@ -861,6 +869,22 @@ definitions:
...
+    type: object
+    description: The associated approval groups
```
bzwei: `description: Parameters to create a workflow`

app/controllers/api/v0/admins_controller.rb
hsong-rh: Yes, `exception_handler` is the overall handler. Here when the input group ids are invalid, we want to return error code 422, rather than 404 of RecordNotFound.

public/doc/swagger-2.yaml
hsong-rh: @bzwei Changed based on your comments.

spec/requests/workflows_spec.rb
```diff
@@ -76,7 +76,11 @@
 
   # Test suite for POST /templates/:template_id/workflows
   describe 'POST /templates/:template_id/workflows' do
-    let(:valid_attributes) { { :name => 'Visit Narnia', :done => false } }
+    let(:group_100) { create(:group, :id => 100) }
```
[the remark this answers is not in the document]
hsong-rh: Want to make sure the next invalid test case does use none exist group ids.

app/services/workflow_create_service.rb
```diff
@@ -0,0 +1,13 @@
+class WorkflowCreateService
+  attr_accessor :template
+  attr_accessor :groups
+
+  def initialize(template_id)
+    self.template = Template.find(template_id)
+  end
+
+  def create(options)
+    Group.find(options[:group_ids])
```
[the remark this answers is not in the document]
hsong-rh: No, next `create` call will create the groups associations. The purpose calling `find` here is to check if groups exist or not.

PR #38, opened 2019-01-15
Add User and its related endpoints

hsong-rh:
This PR introduced a new User model and corresponding endpoints for it.


config/routes.rb
```diff
@@ -28,23 +28,32 @@ def add_swagger_route http_method, path, opts = {}
   end
 
   add_swagger_route 'POST', '/stages/{stage_id}/actions', controller_name: 'admins', constraint_name: 'adminsConstraint', action_name: 'add_action'
+  add_swagger_route 'POST', '/approvers/{approver_id}/actions', controller_name: 'admins', constraint_name: 'adminsConstraint', action_name: 'add_action_by_approver_id'
```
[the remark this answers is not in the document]
bzwei: consider to have endpoint POST `/requests/{request_id}/actions` where we can help to diagnose the stage based on approver's identity.

app/controllers/api/v0/admins_controller.rb
```diff
@@ -36,6 +57,27 @@ def fetch_groups
...
+      end
+
+      def fetch_requests_by_approver_id
+        approver = Approver.find(params.require(:id))
+        requests = approver.requests
+
+        json_response(requests)
+      end
+
+      def fetch_stages_by_approver_id
```
bzwei: We may not need an endpoint to list all stages by an approver. Listing of requests is enough.

app/controllers/api/v0/admins_controller.rb
```diff
@@ -24,6 +29,22 @@ def add_workflow
         json_response({ :message => e.message }, :unprocessable_entity)
       end
 
+      def add_action_by_approver_id
```
bzwei: Not need if we agree do not provide an endpoint for it.

app/models/approver.rb
```diff
@@ -0,0 +1,16 @@
+class Approver < ApplicationRecord
+  acts_as_tenant(:tenant)
+
+  has_many :approvergroups, :dependent => :destroy
+  has_many :groups, :through => :approvergroups
+  validates :username, :presence => true
```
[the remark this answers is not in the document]
hsong-rh: `usename` refers to the user info provided in the header. So far it's not used in any other places.

config/routes.rb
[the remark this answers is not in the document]
hsong-rh: This is based on our discussion: approver is enabled to create an `action` to response the request. For `/requests/{request_id}/actions`, do you want to add it as a new one, or to replace the current `/approvers/{approver_id}/actions`?

app/controllers/api/v0/admins_controller.rb
```diff
@@ -24,6 +29,21 @@ def add_workflow
         json_response({ :message => e.message }, :unprocessable_entity)
       end
 
+      def add_action_by_request_id
+        # TODO
+      end
+
+      def fetch_approver_by_id
```
bzwei: do we have `fetch_approvers_by_group_id`?

config/routes.rb
```diff
@@ -28,13 +28,19 @@ def add_swagger_route http_method, path, opts = {}
...
+  add_swagger_route 'POST', '/approvers', controller_name: 'admins', constraint_name: 'adminsConstraint', action_name: 'add_approver'
   add_swagger_route 'POST', '/groups', controller_name: 'admins', constraint_name: 'adminsConstraint', action_name: 'add_group'
   add_swagger_route 'POST', '/templates/{template_id}/workflows', controller_name: 'admins', constraint_name: 'adminsConstraint', action_name: 'add_workflow'
   add_swagger_route 'GET', '/actions/{id}', controller_name: 'admins', constraint_name: 'adminsConstraint', action_name: 'fetch_action_by_id'
   add_swagger_route 'GET', '/actions', controller_name: 'admins', constraint_name: 'adminsConstraint', action_name: 'fetch_actions'
+  add_swagger_route 'GET', '/approvers/{id}', controller_name: 'admins', constraint_name: 'adminsConstraint', action_name: 'fetch_approver_by_id'
+  add_swagger_route 'GET', '/approvers', controller_name: 'admins', constraint_name: 'adminsConstraint', action_name: 'fetch_approvers'
   add_swagger_route 'GET', '/groups/{id}', controller_name: 'admins', constraint_name: 'adminsConstraint', action_name: 'fetch_group_by_id'
```
bzwei: we should have `GET '/groups/{id}/approvers'`

bzwei: What are the endpoints to add/remove an approver to/from a group?

hsong-rh: Removing an approver from group can share the endpoint of `DELETE '/approvers/{id}'`, because `approver_id` is required in deletion. Instead, we may need an endpoint to delete all approvers from a certain group.

bzwei: We were talking about join/withdraw approvers from a group. It is resolved now.

spec/requests/groups_spec.rb
```diff
@@ -85,4 +95,95 @@
...
+      expect(response).to have_http_status(200)
+    end
+
+    it 'returns number of approvers' do
+      group = Group.find(id)
+      expect(group.approvers.count).to eq(3)
+    end
+  end
+
+  describe 'post /groups/:id/approvers' do
```
bzwei: change endpoint `POST /groups/:id` with body like
```
{
  "operation":"join_approvers",
  "parameters":{
    "approver_ids":[1,2,3]
  }
}
```

public/doc/swagger-2.yaml
```diff
@@ -230,6 +433,39 @@ paths:
           $ref: '#/responses/404NotFoundResponse'
         '422':
           $ref: '#/responses/422UnprocessableEntityResponse'
+    post:
+      tags:
+        - admins
+      summary: Operate (join or withdraw) a list of approvers to a given group
```
bzwei: My intention is that this endpoint to for any operation to a group. Currently supported operations are join and withdraw. We can expand operations in the future.

public/doc/swagger-2.yaml
```diff
@@ -230,6 +433,39 @@ paths:
...
+      description: Approvers join in or withdraw from a given group
+      operationId: operateApproversByGroupId
```
bzwei: method name maybe `groupOperation`

public/doc/swagger-2.yaml
```diff
@@ -230,6 +433,39 @@ paths:
...
+      produces:
+        - application/json
+      parameters:
+        - $ref: '#/parameters/id'
+        - in: body
+          name: body
+          description: Parameters need to operate approvers
+          required: true
+          schema:
+            $ref: '#/definitions/GroupApproversParameters'
```
bzwei: schema name maybe `GroupOperationParameters`

public/doc/swagger-2.yaml
```diff
@@ -742,22 +1067,36 @@ definitions:
...
+  GroupApproversParameters:
+    type: object
+    description: Parameters need to operate approvers to a given group
+    required:
+      - operation
+      - approver_ids
+    properties:
+      operation:
+        $ref: '#/definitions/GroupOperation'
+      approver_ids:
```
bzwei:
```
parameters:
  type: object
```
example: 
```
  "parameters": {
    "approver_ids":[]
  }
```

public/doc/swagger-2.yaml
```diff
@@ -742,22 +1067,36 @@ definitions:
...
-        description: Contact type. It may be email, slack, team, etc. So far only email is supported.
-        example: email
-      contact_setting:
-        type: object
-        description: JSON object with contact Setting
+      approver_ids:
+        type: array
+        items:
+          type: integer
+  GroupApproversParameters:
```
bzwei: `GroupOperationParameters`

public/doc/swagger-2.yaml
```diff
@@ -742,22 +1067,36 @@ definitions:
...
+    type: object
+    description: Parameters need to operate approvers to a given group
```
bzwei: `description: Operation type and parameters to a given group`

public/doc/swagger-2.yaml
```diff
@@ -742,22 +1067,36 @@ definitions:
...
+    required:
+      - operation
+      - approver_ids
```
bzwei: `- parameters`

bzwei:
@hsong-rh please address rubcop warnings.

app/controllers/api/v0/admins_controller.rb
```diff
@@ -81,6 +120,20 @@ def fetch_workflows
         json_response(workflows)
       end
 
+      def group_operation
+        group = Group.find(params.require(:id))
+        group.send(group_params.require(:operation), group_params.require(:approver_ids))
```
bzwei: will this work? `:approver_id` is not a direct child of `group_params`. You should introduce a service class.

spec/requests/groups_spec.rb
```diff
@@ -85,4 +94,95 @@
...
+      group = Group.find(id)
+      expect(group.approvers.count).to eq(3)
+    end
+  end
+
+  describe 'post /groups/:id' do
+    before { post "#{api_version}/groups/#{id}", :params => attributes, :headers => admin_encode_key }
+
+    context 'when new approvers join in group' do
+      let(:attributes) { { :operation => 'join_approvers', :approver_ids => [other_approvers.first.id, other_approvers.last.id] } }
```
bzwei: you need to update this spec.

app/controllers/api/v0/admins_controller.rb
hsong-rh: `parameters` is the keyword in controller. We need to choose  another word to replace it.

bzwei: arguments?

bzwei: or params?

config/routes.rb
```diff
@@ -1,18 +1,15 @@
...
-Insights Service Approval APIs
-
-APIs to query approval service
-
-OpenAPI spec version: 1.0.0
-Contact: support@redhat.com
-Generated by: https://github.com/swagger-api/swagger-codegen.git
-
-=end
+# Insights Service Approval APIs
```
bzwei: remove this block

bzwei:
@gmcculloug @mkanoor please review

public/doc/swagger-2.yaml
```diff
@@ -12,8 +12,8 @@ host: 'localhost:3000'
 tags:
   - name: admins
     description: Admin-only APIs
-  - name: approvers
-    description: Approvers APIs
+  - name: users
```
bzwei: You can't have two `users` tags. Should the regular user be `requesters` or `guests`?

public/doc/swagger-2.yaml
```diff
@@ -230,10 +433,43 @@ paths:
...
+      operationId: groupOperation
+      consumes:
+        - application/json
+      produces:
+        - application/json
+      parameters:
+        - $ref: '#/parameters/id'
+        - in: body
+          name: body
+          description: Parameters need to operate users
```
bzwei: Parameters need to operate the group

gmcculloug:
@hsong-rh Please update the PR title/description to reflect the change from Approver to User.

public/doc/swagger-2.yaml
```diff
@@ -728,6 +999,31 @@ definitions:
         format: date-time
         description: Timestamp of creation
         readOnly: true
+  Approver:
+    type: object
+    description: >-
+      Approver with all personal information.
+    required:
+      - username
```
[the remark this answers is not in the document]
mkanoor: @bzwei We would need the users last name and first name to craft emails and payloads.

app/controllers/api/v0/admins_controller.rb
mkanoor: @bzwei I think instead of stages we should be calling this levels, since the feature is called multi   level approval. The stage means different steps in a single process.

app/services/group_operation_service.rb
```diff
@@ -0,0 +1,19 @@
+class GroupOperationService
```
mkanoor: @hsong-rh 
Do you want to follow the pattern in service portal?
https://github.com/ManageIQ/service_portal-api/blob/master/app/services/service_catalog/service_plans.rb

bzwei: @hsong-rh was following the existing pattern in approval-api. We have pattern like `GroupOperationService#operate` or `WorkflowCreateService#create`. Even though the name appears a little redundant, we feel it is good for readability. We know immediately it is a service class, and method tells what the method intents to do. We don't limit the method name to a fixed one like `process`. It provides us a little more flexibility. For example the same service class may be able to accommodate two slightly different flavors, such as `create` and `create_if`.

We should create another PR to change the coding style for all service methods if we do need to refactor them.

mkanoor:
@hsong-rh 
The Operation and GroupOperation in the swagger.yaml doesn't have object and properties and when we generate the client code these objects come out empty.
https://github.com/ManageIQ/approval-api/blob/9d5b330927d7417f191412ef964379e738e31bf3/public/doc/swagger-2.yaml#L762
https://github.com/ManageIQ/approval-api/blob/9d5b330927d7417f191412ef964379e738e31bf3/public/doc/swagger-2.yaml#L732
https://github.com/ManageIQ/approval-api/blob/9d5b330927d7417f191412ef964379e738e31bf3/public/doc/swagger-2.yaml#L830

mkanoor:
@hsong-rh @bzwei 
Shouldn't the operation be part of the URL? so you could have /request/{id}/approve or /request/{id}/deny?

hsong-rh:
@mkanoor Both Operation and GroupOperation are internal used by service. They don't expose to client side. The request's operation is handled indirectly by Action on Stage, not via endpoints.

PR #39, opened 2019-01-16
Validate whether an action is allowed at current statge state

bzwei:
Allowed state transition through action:
pending -> notify -> notified
notified -> approve/deny -> finished
pending -> skip -> skipped
memo action is always allowed.

Raise an error if not one of above cases.

### DIFF
--- a/spec/services/action_create_service_spec.rb
+++ b/spec/services/action_create_service_spec.rb
@@ -16,7 +16,7 @@ RSpec.describe ActionCreateService do
 
   context 'notify operation' do
     it 'updates stage and request' do
-      action = svc1.create(:operation => Action::NOTIFY_OPERATION, :processed_by => 'system')
+      action = svc1.create('operation' => Action::NOTIFY_OPERATION, 'processed_by' => 'system')
       stage1.reload
       request.reload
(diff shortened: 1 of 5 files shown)


spec/services/action_create_service_spec.rb
```diff
@@ -27,7 +27,8 @@
 
   context 'approve/deny operation' do
     it 'updates current stage' do
-      action = svc1.create(:operation => Action::APPROVE_OPERATION, :processed_by => 'man')
+      stage1.update_attributes(:state => Stage::NOTIFIED_STATE)
+      action = svc1.create('operation' => Action::APPROVE_OPERATION, 'processed_by' => 'man')
```
gmcculloug: @bzwei Can you explain why we seem to be switching calls from symbols to string, but adding new line that are using symbols?  This pattern repeats in this spec.  And the changes to `app/services/request_create_service.rb` are switching over to strings.  I'm not if there is a purpose or why it is not consistent.

bzwei: `update_attributes` as an AR method can take either symbol or string. 
`svc1.create` is not an AR method. It is normally called from the controller where it receives REST call with JSON body.

PR #47, opened 2019-02-19
Convert api from swagger 2.0 to openapi 3.0

hsong-rh:
Convert to OpenAPI 3.0


hsong-rh:

bzwei:
@hsong-rh were you able to generate an openapi client and connect to the api server?

hsong-rh:
@bzwei No, I tested it by querying openapi.json directly via local server. Will use openapi generator to test it now.

hsong-rh:
@bzwei I tested with openapi generated client, it works as expected.

PR #48, opened 2019-02-20
Support major version forwarding

hsong-rh:
Support major version forwarding.

Reference: https://github.com/ManageIQ/service_portal-api/pull/109

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -9,6 +9,7 @@ Rails.application.routes.draw do
 
     full_path = path.gsub(/{(.*?)}/, ':\1')
     scope :as => :api, :module => "api", :path => prefix do
+      match "/v0/*path", :via => [:delete, :get, :options, :patch, :post], :to => redirect(:path => "/#{prefix}/v0.0/%{path}", :only_path => true)
       namespace :v0x0, :path => "v0.0" do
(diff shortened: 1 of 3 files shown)


config/routes.rb
```diff
@@ -9,6 +9,7 @@ def add_swagger_route(http_method, path, opts = {})
 
     full_path = path.gsub(/{(.*?)}/, ':\1')
     scope :as => :api, :module => "api", :path => prefix do
+      match "/v0/*path", :via => [:delete, :get, :options, :patch, :post], :to => redirect(:path => "/#{prefix}/v0.0/%{path}", :only_path => true)
```
gmcculloug: I think the idea was to redirecting to the latest minor version, redirecting to `0.0` was meant to be a temporary change.  cc @syncrou

syncrou: Correct, we will be defaulting to v0.1.0 on service_portal after the resource routes are merged ( as a separate PR. )

hsong-rh: Ok, `v0.0` is the only available version so far in approval service. Will update to `v0.1.0` when it's needed.

bzwei: Isn't it `0.0.1` with last PR(#47) merged?

hsong-rh: @bzwei `0.0.1` is the version for openapi.json. I think we only care the major and minor versions so far.

PR #49, opened 2019-02-25
Fixed  type and format for notified_at

lgalis:
Fixed  type and format for notified_at

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,5 +11,6 @@
 /log/*.log
 /pkg/
 /spec/reports/
+.idea/*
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -11,5 +11,6 @@
 /log/*.log
 /pkg/
 /spec/reports/
+.idea/*
```
gmcculloug: Did you mean to include this?  Seems like it could go into your local `.gitignore_global` file

lgalis: Removed

PR #50, opened 2019-02-25
Separates controllers based on resources

hsong-rh:
Separates controllers based on resources, applies resource based routes to the `api/v0.1/` version of the API. All old routes are kept for backwards compatibility on `/api/v0.0/`


hsong-rh:
@bzwei @mkanoor @syncrou Please review.

PR #52, opened 2019-03-01
Fix openapi.json to set allOf correctly

hsong-rh:
Fix `allOf` inheritance relationship in openapi json file

### DIFF
--- a/public/approval/v0.0.1/openapi.json
+++ b/public/approval/v0.0.1/openapi.json
@@ -1774,25 +1774,29 @@
                 }
             },
             "ActionOut": {
+                "description": "Output parameters for Action object",
                 "allOf": [
                     {
                         "$ref": "#/components/schemas/ActionIn"
-                    }
-                ],
-                "properties": {
-                    "id": {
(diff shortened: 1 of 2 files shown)


public/approval/v0.0.1/openapi.json
```diff
@@ -1817,13 +1821,17 @@
                 "allOf": [
                     {
                         "$ref": "#/components/schemas/UserIn"
+                    },
+                    {
+                        "type": "object",
+                        "description": "Onput parameters for User object",
```
bzwei: In https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md I saw `"description"` is at top level parallel with `"offOf"`.

hsong-rh: Yes, the `description` will be automated skipped when creating client side docs.

bzwei: let's move to top for better readability.

hsong-rh: Changed.

PR #53, opened 2019-03-05
Adds pagination support

hsong-rh:
Added pagination support in v0.1.0


hsong-rh:

app/controllers/api/v0x1/mixins/index_mixin.rb
```diff
@@ -0,0 +1,27 @@
...
+            ref_schema = {relation.model.tagging_relation_name => :tag}
+
+            relation = relation.includes(ref_schema).references(ref_schema)
+          end
+          relation
+        end
+
+        def collection(base_query)
+          resp = ManageIQ::API::Common::PaginatedResponse.new(
+            :base_query => base_query,
```
syncrou: Why not `:base_query => scoped(base_query),` and then call `collection(< base_query>)` from `index` methods?

hsong-rh: Good point, changed.

PR #54, opened 2019-03-06
Set current request context and set tenant from header

bzwei:
In the `around_action` block set the current request and tenant for all controller calls.


spec/support/request_spec_helper.rb
```diff
@@ -5,7 +5,46 @@ def json
     JSON.parse(response.body)
   end
 
-  def version(v = 'v0.0')
-    "/api/#{v}"
+  def version(ver = 'v0.0')
```
syncrou: @bzwei if you add the support `Dir` from the common gem, you don't need to duplicate this support helper. https://github.com/ManageIQ/service_portal-api/pull/135/files#diff-93830fa29d616f7c87903d08b5b1b29a

hsong-rh: @syncrou For a more flexible and readable purpose, it's better to keep its own spec help methods.

app/controllers/application_controller.rb
```diff
@@ -1,4 +1,24 @@
 class ApplicationController < ActionController::API
   include Response
   include ExceptionHandler
+
+  around_action :with_current_request
+
+  private
+
+  def with_current_request
```
syncrou: @bzwei do you see no need to allow for an env variable to bypass_tenancy?

app/controllers/application_controller.rb
```diff
@@ -1,4 +1,24 @@
...
+    ManageIQ::API::Common::Request.with_request(request) do |current|
+      begin
+        ActsAsTenant.with_tenant(current_tenant(current.user)) { yield }
+      rescue Exceptions::NoTenantError
+        json_response({ :message => 'Unauthorized' }, :unauthorized)
+      end
```
mkanoor: @bzwei Is this missing an ensure to unset the thread local storage like here

https://github.com/ManageIQ/manageiq-api-common/blob/6d242a7fded6935557b769658480e364705ca6f9/lib/manageiq/api/common/request.rb#L30

mkanoor: @bzwei 
ensure
  ActsAsTenant.current_tenant = nil

mkanoor: Actually they are already doing this
https://www.rubydoc.info/gems/acts_as_tenant/0.3.9/ActsAsTenant.with_tenant

spec/models/tenant_spec.rb
```diff
@@ -24,6 +24,10 @@
         expect(template.tenant_id).to eq tenant.id
       end
     end
+
+    after do
+      ActsAsTenant.current_tenant = nil
```
mkanoor: @bzwei If using the with_tenant you don't need this
https://www.rubydoc.info/gems/acts_as_tenant/0.3.9/ActsAsTenant.with_tenant

hsong-rh: @mkanoor This is the part we mentioned during standup. This `after` will cleanup leftover tenants.

app/controllers/application_controller.rb
bzwei: no for now. We will follow your example if we feel it is needed for test/development

spec/models/tenant_spec.rb
bzwei: Yes, we should use `with_tenant`. Fixed.

PR #58, opened 2019-03-11
Change date and id in right format

hsong-rh:
Change attributes into right format.


app/models/action.rb
```diff
@@ -8,7 +8,18 @@ class Action < ApplicationRecord
...
   validates :operation,    :inclusion => { :in => OPERATIONS }
   validates :processed_by, :presence  => true
   belongs_to :stage, :inverse_of => :actions
+
+  def as_json(_options = {})
+    attributes.slice(*NON_DATE_ATTRIBUTES).tap do |hash|
+      DATE_ATTRIBUTES.each do |attr|
+        hash[attr] = send(attr.to_sym).iso8601 if send(attr.to_sym)
+      end
+    end.merge(:id => id.to_s)
```
gmcculloug: Constants are the wrong way to go here.

You can check for the a date-time field a few ways:

1) Check the type of column:
```ruby
pry(#<Service>):1> column_for_attribute('created_at').type
=> :datetime
```

2) Checking the type of value:
```ruby
pry(#<Service>):1> attribute('created_at').kind_of?(Time)
=> true
```

@syncrou @mkanoor Do we have `as_json` logic anywhere else that we should be sharing here?  Seems we are just converting date fields, making `id` a string and dropping `tenant_id`.

Also, should `as_json` be using symbols?  It feels like the whole thing could be a small method build on `attributes.except('tenant_id')`.

And the duplication makes me think mixin.

mkanoor: @gmcculloug 
We have something in topology that needs to be exposed to all services where we read the openapi.json file and look at the response object and convert it to the valid strings. The  list would be defined in the openapi.json

https://github.com/ManageIQ/topological_inventory-api/blob/3300b0262aa686e0e3c29c77e54235b5c30425b5/lib/open_api/serializer.rb#L3

I think @syncrou had started working on it at some point I am hoping it will get integrated into  manageiq-api-common

gmcculloug: @syncrou Any thoughts here?  Should we copy the Topology pattern here until we can get it into common code?

hsong-rh: @gmcculloug I just made changes based on your previous comments.

app/models/concerns/convertable.rb
```diff
@@ -0,0 +1,12 @@
+module Convertable
+  extend ActiveSupport::Concern
+
+  def convert_date_id(attributes_hash)
+    attributes_hash.tap do |hash|
+      attributes_hash.each_key do |attr_key|
+        hash[attr_key] = send(attr_key.to_sym).iso8601 if attribute(attr_key).kind_of?(Time) && send(attr_key.to_sym)
```
bzwei: `#send` can take a string. You don't need to convert to symbol.

lindgrenj6: we could also use `respond_to?(attr_key)` instead of looking for nil coming back from `send`

hsong-rh: @lindgrenj6 @bzwei Changed.

gmcculloug:
@hsong-rh See https://github.com/ManageIQ/catalog-api/pull/152 as a possible alternative approach.

hsong-rh:
@syncrou @bzwei Changed to @gmcculloug's approach. Please review.

hsong-rh:
@gmcculloug Spec tests are added.

spec/requests/v0.1/actions_spec.rb
```diff
@@ -22,7 +23,8 @@
     context 'when the record exists' do
       it 'returns the action' do
         expect(json).not_to be_empty
-        expect(json['id']).to eq(id)
+        expect(json['id']).to eq(id.to_s)
+        expect(json['created_at']).to eq(created_at.iso8601.to_s)
```
gmcculloug: Minor but, I would drop the extract `let` blocks for the specific properties and only do the lookup once in the test with `action = actions.first` and then `expect` would test each property:
```
action = actions.first

expect(json['id']).to eq(action.id.to_s)
expect(json['created_at']).to eq(action.created_at.iso8601)
```

Also, `iso8601` returns a string so you can drop the `to_s`.

hsong-rh: @gmcculloug Changed accordingly.

PR #59, opened 2019-03-11
Removed unused version 0x0

hsong-rh:
Cleanup outdated version 0x0


hsong-rh:
@bzwei Please review.

PR #61, opened 2019-03-13
Renamed operationId to avoid duplicates

hsong-rh:
This PR fixed two issues:
1. Removed duplicated operationId: 
    https://projects.engineering.redhat.com/browse/SSP-201
2. Added SUCCESS response for workflow's DELETE/PATCH actions: 
    https://projects.engineering.redhat.com/browse/SSP-212

### DIFF
--- a/public/approval/v0.1.0/openapi.json
+++ b/public/approval/v0.1.0/openapi.json
@@ -46,9 +46,6 @@
                 "parameters": [
                     {
                         "$ref": "#/components/parameters/stage_id"
-                    },
-                    {
-                        "$ref": "#/components/parameters/identity"
                     }
                 ],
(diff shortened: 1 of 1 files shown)


hsong-rh:
@bzwei Please review.

bzwei:
@hsong-rh Why do we need to allow client to set `x-rh-identity`? If it is for developer testing purpose we should consider alternatives.

hsong-rh:
@bzwei Yes, it's for our own client side test script. We need to set header info to make sure all endpoints working properly. The parameter is optional, means no changes in API side if not needed.

hsong-rh:
@bzwei As discussed, I removed optional header parameter part.

PR #64, opened 2019-03-15
Add kie_client in lib

bzwei:


lib/rbac/service.rb
```diff
@@ -51,8 +51,8 @@ def self.paginate(obj, method, pagination_options, *method_args)
     private_class_method def self.dev_credentials(config)
       # Set up user/pass for basic auth if we're in dev and they exist.
       if Rails.env.development?
-        config.username = ENV['DEV_USERNAME'] || raise("Empty ENV variable: DEV_USERNAME")
-        config.password = ENV['DEV_PASSWORD'] || raise("Empty ENV variable: DEV_PASSWORD")
+        config.username = ENV['RBAC_USERNAME'] || raise("Empty ENV variable: RBAC_USERNAME")
+        config.password = ENV['RBAC_PASSWORD'] || raise("Empty ENV variable: RBAC_PASSWORD")
```
gmcculloug: @bzwei Why the change to `RBAC_USERNAME` and `RBAC_PASSWORD`?  Wouldn't the old names work just as well and be consistent with the catalog api?  Plus, I would expect the `dev_credentials` method to use `DEV` environment variables and not `RBAC`.  Realistically, these are for authenticating to the environment where the services run, not the individual services themselves, so the more generic name would be better.

Even if we want this change does it belong in this `kie_client` specific PR?  I do not see the relationship.

bzwei: I was trying to normalize the naming convention. Since we now need multiple sets of user/password env, `DEV_USERNAME` dedicated to RBAC solely does not distinguish itself from others. Thus I prefer `RBAC_USERNAME` and `KIESERVER_USERNAME`.

Gemfile
```diff
@@ -17,6 +17,7 @@ gem 'swagger_ui_engine'
 
 gem 'manageiq-api-common', :git => 'https://github.com/ManageIQ/manageiq-api-common', :branch => 'master'
 gem 'rbac-api-client', :git => "https://github.com/mkanoor/rbac_api_client", :branch => "master"
+gem 'kie_client', :git => "https://github.com/bzwei/kie_client", :branch => "master"
```
syncrou: @gmcculloug I know we're trying to move everything to the `ManageIQ` repo. Should we try to get gems  created in that repo first?

lib/rbac/service.rb
syncrou: @bzwei is there a case where the username/password could be different between the services when accessing them from a development environment?

bzwei:
will be replaced with another PR.

PR #67, opened 2019-03-19
Start or signal a business process through kie_client

bzwei:
When the `process_setting` and `signal_setting` are configured for a template to connect to an external jbpm kieserver, the code now creates a new process instance when a request is created, and signals the instance when a stage is completed.

### DIFF
--- a/app/services/jbpm_process_service.rb
+++ b/app/services/jbpm_process_service.rb
@@ -1,0 +1,37 @@
+class JbpmProcessService
+  attr_accessor :request
+
+  def initialize(request)
+    self.request = request
+  end
+
+  def start
+    options = template.process_setting
+    Kie::Service.call(KieClient::ProcessInstancesBPMApi, options) do |bpm|
+      bpm.start_process(options['container_id'], options['process_id'], :body => process_options)
+    end
+  end
+
+  def signal(decision)
+    options = template.signal_setting
+    Kie::Service.call(KieClient::ProcessInstancesBPMApi, options) do |bpm|
+      bpm.signal_process_instance(options['container_id'], request.process_ref, options['signal_name'], :body => signal_options(decision))
+    end
+  end
+
+  private
+
+  def template
+    request.workflow.template
+  end
+
+  def process_options
+    # TODO: more options
+    request.content
+  end
+
+  def signal_options(decision)
+    # TODO: more options
+    {'decision' => decision}
+  end
+end
--- a/app/services/request_create_service.rb
+++ b/app/services/request_create_service.rb
@@ -22,9 +22,11 @@ class RequestCreateService
     )
     Request.create!(create_options).tap do |request|
       if default_approve? || auto_approve?
-        start_approval_process(request)
+        start_internal_approval_process(request)
       elsif !workflow.external_processing?
         start_first_stage(request)
+      else
+        start_external_approval_process(request)
       end
     end
   end
@@ -39,7 +41,7 @@ class RequestCreateService
     ENV['AUTO_APPROVAL'] && ENV['AUTO_APPROVAL'] != 'n'
   end
 
-  def start_approval_process(request)
+  def start_internal_approval_process(request)
     Thread.new do
       default_approve? ? default_approve(request) : auto_approve(request)
     end
@@ -48,6 +50,7 @@ class RequestCreateService
   def auto_approve(request)
     sleep_time = ENV['AUTO_APPROVAL_INTERVAL'].to_f
 
+    request_started(request)
     request.stages.each { |stage| group_auto_approve(stage, sleep_time) }
   end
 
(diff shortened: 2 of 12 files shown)


lib/kie/service.rb
```diff
@@ -0,0 +1,27 @@
...
+      Rails.logger.error("KieClient::ApiError #{err.message} ")
+      raise Exceptions::KieError, err.message
+    end
+
+    private_class_method def self.setup(options)
+      KieClient.configure do |config|
+        config.host     = options['host'] || 'localhost'
+        config.scheme   = options['host'].try(:scheme) || 'http'
+        config.username = options['username']
+        config.password = options['password']
```
bzwei: This is a temporary solution. We will use manageiq-api-common to manage passwords in a follow-up PR.

app/services/request_create_service.rb
```diff
@@ -88,9 +91,18 @@ def request_finished(request)
...
+    request_started(request)
     ActionCreateService.new(request.stages.first.id).create(
       :operation    => Action::NOTIFY_OPERATION,
       :processed_by => 'system'
     )
   end
+
+  def start_external_approval_process(request)
+    template = request.workflow.template
+    processor_class = "#{template.process_setting['processor_type']}_process_service".classify.constantize
```
hsong-rh: What's the benefit we use such a way here and next to initialize `jbpm_process_service`? This will require both `process_setting `and `signal_setting` (in `stage_update_service`) to have attribute `processor_type`. What happens if it's not been defined?

app/services/jbpm_process_service.rb
```diff
@@ -0,0 +1,37 @@
...
+  end
+
+  def start
+    options = template.process_setting
+    Kie::Service.call(KieClient::ProcessInstancesBPMApi, options) do |bpm|
+      bpm.start_process(options['container'], options['process'], :body => process_options)
+    end
+  end
+
+  def signal(decision)
```
hsong-rh: In our case, we just send `signal` and let approval process go, may not need `signal_option` here, or keep it but provide a default value.

app/services/jbpm_process_service.rb
```diff
@@ -0,0 +1,37 @@
...
+  attr_accessor :request
+
+  def initialize(request)
+    self.request = request
+  end
+
+  def start
+    options = template.process_setting
+    Kie::Service.call(KieClient::ProcessInstancesBPMApi, options) do |bpm|
+      bpm.start_process(options['container'], options['process'], :body => process_options)
```
hsong-rh: For clarification, it's better to use the same parameter name as defined in Kie server doc, like
`bpm.start_process(options['container_id'], options['process_id'], :body => process_options)`

app/services/jbpm_process_service.rb
```diff
@@ -0,0 +1,37 @@
...
+    end
+  end
+
+  def signal(decision)
+    options = template.signal_setting
+    Kie::Service.call(KieClient::ProcessInstancesBPMApi, options) do |bpm|
+      bpm.signal_process_instance(options['container'], request.process_ref, options['signal'], :body => signal_options(decision))
```
hsong-rh: Same reason as before,
`bpm.signal_process_instance(options['container_id'], request.process_ref, options['signal_name'], :body => signal_options(decision))`

app/services/jbpm_process_service.rb
bzwei: We do need to send the decision to bpm. BPM will decide whether to skip the next group.

app/services/request_create_service.rb
bzwei: We talked about supporting of external approval processors. The `processor_type` allows us to expand such capacity.

hsong-rh: 👍

app/services/jbpm_process_service.rb
hsong-rh: 👍

app/services/request_create_service.rb
gmcculloug: Not for this PR, but would be good to move this line for figuring out the `process_class` into a separate method at some point.  Would help with testing when we do support external processors.

app/services/stage_update_service.rb
```diff
@@ -62,10 +55,24 @@ def request_finished(last_decision, last_reason)
...
     next_stage = stage.request.stages.find { |s| s.state == Stage::PENDING_STATE }
     return unless next_stage
 
     operation = decision == Stage::DENIED_STATUS ? Action::SKIP_OPERATION : Action::NOTIFY_OPERATION
     ActionCreateService.new(next_stage.id).create(:operation => operation, :processed_by => 'system')
   end
+
+  def signal_external_approval_process(decision)
+    template = stage.request.workflow.template
+    processor_class = "#{template.signal_setting['processor_type']}_process_service".classify.constantize
```
gmcculloug: Same here.  Slightly different lookup, but would be nice as a separate method at some point.

PR #72, opened 2019-03-28
Add version into prefix path

hsong-rh:


hsong-rh:
@bzwei Please review and merge.

PR #73, opened 2019-03-28
Changed version to v1

hsong-rh:
Change API version from V0x1 to V1x0


config/initializers/swagger_ui_engine.rb
```diff
@@ -2,7 +2,6 @@
 
 SwaggerUiEngine.configure do |config|
   config.swagger_url = {
-    v0_0_1: '/approval/v0.0.1/openapi.json',
-    v0_1_0: '/approval/v0.1.0/openapi.json',
+    v1_0_0: '/approval/v1.0/openapi.json',
```
syncrou: My understanding was that `SwaggerUIEngine` doesn't work with `openapi.json`.

spec/requests/v1.0/actions_spec.rb
```diff
@@ -13,7 +13,7 @@
   let!(:actions) { create_list(:action, 10, :stage_id => stage.id, :tenant_id => tenant.id) }
   let(:id) { actions.first.id }
 
-  let(:api_version) { version('v0.1') }
+  let(:api_version) { version('v1.0') }
```
syncrou: If the `version` method default is modified to v1.0 these calls can be simplied
```
def version(ver = 'v0.0')
    "/api/#{ver}"
  end
```

spec/requests/v1.0/requests_spec.rb
```diff
@@ -14,7 +14,7 @@
   let!(:requests_with_same_state) { create_list(:request, 2, :state => 'notified', :workflow_id => workflow.id, :tenant_id => tenant.id) }
   let!(:requests_with_same_decision) { create_list(:request, 2, :decision => 'approved', :workflow_id => workflow.id, :tenant_id => tenant.id) }
```
syncrou: I noticed `spec/requests/` uses a `v1.0` directory structure and `spec/controllers/` uses a `v1` directory structure. Is there a reason for the difference?

hsong-rh:
@syncrou Changed accordingly, please review again

PR #74, opened 2019-03-28
Add a random id column to stages table

bzwei:
It is used by external limit access without login. It gets cleared once the stage is in finished state.

### DIFF
--- a/app/services/request_create_service.rb
+++ b/app/services/request_create_service.rb
@@ -1,4 +1,6 @@
 class RequestCreateService
+  require 'securerandom'
+
   attr_accessor :workflow
 
   def initialize(workflow_id)
@@ -8,9 +10,10 @@ class RequestCreateService
   def create(options)
     stages = workflow.group_refs.collect do |group_ref|
       Stage.new(
-        :group_ref => group_ref,
-        :state     => Stage::PENDING_STATE,
-        :decision  => Stage::UNDECIDED_STATUS,
+        :group_ref         => group_ref,
+        :state             => Stage::PENDING_STATE,
+        :decision          => Stage::UNDECIDED_STATUS,
+        :random_access_key => SecureRandom.hex(16)
       )
     end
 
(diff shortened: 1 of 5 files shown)


db/migrate/20190329151023_add_rid_to_stages.rb
```diff
@@ -0,0 +1,7 @@
+class AddRidToStages < ActiveRecord::Migration[5.1]
+  def change
+    add_column :stages, :rid, :string
```
gmcculloug: This name is not meaningful and will be very confusion to others.  Maybe `secure_access_key` or `randon_access_key`.  The name should be easily identifiable.

app/services/request_create_service.rb
```diff
@@ -50,25 +53,18 @@ def start_internal_approval_process(request)
...
-    )
-
     sleep(sleep_time)
-    acs.create(
+    ActionCreateService.new(stage.id).create(
       'operation'    => Action::APPROVE_OPERATION,
       'processed_by' => 'system',
       'comments'     => 'ok'
     )
+    4+5
```
syncrou: Was this an intentional return value or a typo?

app/services/request_create_service.rb
```diff
@@ -8,9 +10,10 @@ def initialize(workflow_id)
   def create(options)
     stages = workflow.group_refs.collect do |group_ref|
       Stage.new(
-        :group_ref => group_ref,
-        :state     => Stage::PENDING_STATE,
-        :decision  => Stage::UNDECIDED_STATUS,
+        :group_ref         => group_ref,
+        :state             => Stage::PENDING_STATE,
+        :decision          => Stage::UNDECIDED_STATUS,
+        :random_access_key => SecureRandom.hex(16)
```
gmcculloug: I realize `hex` is safe for our needs, but did you consider any of the other options.  

Other services that I've seen use this appear to be using `SecureRandom.urlsafe_base64` and I could see extending beyond the default 16-chars, but I think we can adjust that almost anytime.

bzwei: Agree. We could adjust this anytime as needed.

PR #78, opened 2019-03-29
Allow no requester is set in request creation

hsong-rh:
https://github.com/ManageIQ/approval-api/issues/76


bzwei:
This is a temporary solution. It allows a request to be created without populating the requester field for now. We need some more thoughts on how to properly save requester with a request.

gmcculloug:
@mkanoor @syncrou Can this be solved with how we set owner in catalog? 
 https://github.com/ManageIQ/catalog-api/blob/master/app/models/concerns/owner_field.rb

PR #81, opened 2019-04-02
Add context column to Request model

lindgrenj6:
https://projects.engineering.redhat.com/browse/SSP-263

Adding a `context` column to the Request object so we can track how things move through the system. It just stores the `ManageIQ::API::Common::Request.current` object - which is a hash that contains "headers" and "original_url".

### DIFF
--- a/app/models/request.rb
+++ b/app/models/request.rb
@@ -17,4 +17,12 @@ class Request < ApplicationRecord
   scope :decision,  ->(decision)  { where(:decision => decision) }
   scope :state,     ->(state)     { where(:state => state) }
   scope :requester, ->(requester) { where(:requester => requester) }
+
+  after_initialize :set_context
+
+  private
+
+  def set_context
+    self.context = ManageIQ::API::Common::Request.current.to_h
+  end
 end
--- a/db/migrate/20190402165411_add_context_to_request.rb
+++ b/db/migrate/20190402165411_add_context_to_request.rb
@@ -1,0 +1,5 @@
+class AddContextToRequest < ActiveRecord::Migration[5.2]
+  def change
+    add_column :requests, :context, :jsonb
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2019_03_29_151023) do
+ActiveRecord::Schema.define(version: 2019_04_02_165411) do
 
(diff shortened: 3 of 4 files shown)


lindgrenj6:
@miq_bot add_reviewer mkanoor

lindgrenj6:
@miq-bot add_reviewer bzwei

lindgrenj6:
@miq-bot add_reviewer mkanoor

lindgrenj6:

lindgrenj6:
@miq-bot add_reviewer syncrou

bzwei:
We don't have immediate need for this feature. Let's reexamine it after summit. 
If we want to add the context column, we should do the same to actions table too.

mkanoor:
@bzwei Isn't the request connected to the Action, why do we need it in multiple places. Also this feature is needed for debugging purposes, debugging multiple services becomes problematic when we have to go thru multiple log files. Once the logging of the request id is enabled in the manageiq-api-common gem the logging will happen for free if you set the context.

And the logic for setting the context is very straight forward it uses all the code from manageiq-api-common gem. Like shown here
https://github.com/ManageIQ/catalog-api/blob/b2ed5b108af4f514bcf2d2d5f9b8843d3d06af40/app/services/catalog/order_item_transition.rb#L13

bzwei:
@mkanoor approval-api does not need to reconstruct the request in order to do logging. Everything is live with current request from api controller.
 
Does manageiq-api-common gem require a model to have the context column? Shouldn't it always read from current request?

Will we rely on kibana centralized logs or will we actually connect to the database to read the context column to do debugging? Remember we don't expose the context column through the API. If we don't directly connect to the DB to read the context column, there is no big value to store them and use up a lot of storage. For catalog-api it is reasonable because we need to reconstruct the request header.

mkanoor:
@bzwei The serializing of the data is happening in the context column. When we have to resume work later on for a request which might be after couple of days/hours we need to reconstitute the request from the context. Any work that you have in the background needs to log with the x-rh-insights-request-id which will happen seamlessly via the common gem. We understand when you create the Approval Request you have the current_request. What happens when you have to process this request later on based on approvals that come in later, how do we track that. We will have tracking when the request is first created after that there will be no more logging with the x-rh-insights-request-id, how do we trace these things. If you run you code the ones that run in the background not via the controller this context will allow you to set the current request which will be logged correctly by the manageiq-api-common gem

bzwei:
Later the code will run under approver's context. For debugging purpose I see the value to switch to the original requester's context just to add a log.

PR #84, opened 2019-04-04
Seed built-in multi-level approval template to link to internal kieserver

bzwei:
1. Create the encryptions table to store encrypted secrets.
2. Config `process_setting` and `signal_setting` of the build-in multi-level approval template to communicate with the internal kieserver through ENV variables.
3. Encrypt passwords

### DIFF
--- a/app/models/template.rb
+++ b/app/models/template.rb
@@ -5,10 +5,54 @@ class Template < ApplicationRecord
 
   validates :title, :presence => :title
 
+  before_destroy :delete_passwords
+
   def self.seed
     template = find_or_create_by!(:title => 'Basic')
     template.update_attributes(
-      :description => 'A basic approval workflow that supports multi-level approver groups through email notification'
+      :description     => 'A basic approval workflow that supports multi-level approver groups through email notification',
+      :process_setting => seed_process_setting(template.process_setting),
+      :signal_setting  => seed_signal_setting(template.signal_setting)
     )
+  end
+
+  private_class_method def self.seed_process_setting(old_setting)
+    return nil unless ENV['KIE_SERVER_HOST']
+    seed_bpm_setting(old_setting).merge('process_id' => ENV['BPM_BML_PROCESS_ID'])
+  end
+
+  private_class_method def self.seed_signal_setting(old_setting)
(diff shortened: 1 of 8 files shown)


gmcculloug:
@hsong-rh Please review

app/models/template.rb
```diff
@@ -5,10 +5,54 @@ class Template < ApplicationRecord
...
+      else
+        old_password.try(:delete)
+        Encryption.create!(:secret => ENV['KIE_SERVER_PASSWORD']).id
+      end
+
+    {
+      'container_id' => ENV['KIE_CONTAINER_ID'],
+      'password'     => new_password_id,
+      'username'     => ENV['KIE_SERVER_USERNAME'],
+      'host'         => ENV['KIE_SERVER_HOST']
```
hsong-rh: We should also set base_path here. It'a `/kie-server/services/rest`

bzwei: this path is automatically set by the kie-client

app/models/template.rb
```diff
@@ -5,10 +5,54 @@ class Template < ApplicationRecord
...
+    }
+  end
+
+  private
+
+  def delete_passwords
+    process_password_id = process_setting.try(:[], 'password')
+    Encryption.delete(process_password_id) if process_password_id
```
syncrou: Why not use `.destroy!` here as opposed to `.delete`?

app/models/template.rb
```diff
@@ -5,10 +5,54 @@ class Template < ApplicationRecord
...
+
+    signal_password_id = signal_setting.try(:[], 'password')
+    Encryption.delete(signal_password_id) if signal_password_id
```
syncrou: same comment as above.

app/models/template.rb
bzwei: No difference. There is no direct association between `Encryption` and any other models

syncrou: But, if you call `.destroy!` and there is an Exception it will cancel the call back and the error will be bubbled up.  With `.delete` you will not get any feedback and the callback will continue as if nothing was wrong.

PR #86, opened 2019-04-05
Added an open link for approver to process request

hsong-rh:
The link is like: `/api/approval/v#/stageaction`, the views are based on PM's example, as follows:

The request page:
image

The response page:
image


bzwei:
@hsong-rh was just an image scaling matter? The first screenshot the Red Hat icon and Hybrid Cloud Manager looks smaller but Catalog Request and its content look larger.

hsong-rh:
@bzwei Yes, I have to zoomed out in the request page, so all elements can be taken in the screenshot.

app/controllers/api/v1x0/stageaction_controller.rb
```diff
@@ -0,0 +1,85 @@
...
+        set_view_path
+        set_resources
+
+        @approver = params.require(:approver)
+        @stage = Stage.find_by(:random_access_key => params.require(:id))
+
+        if @stage
+          @order = set_order
+        else
+          response.body = "Your request [#{params[:id]}] is either expired or already been processed!"
```
bzwei: replace `already` by `has`

app/controllers/api/v1x0/stageaction/result.html.erb
```diff
@@ -0,0 +1,39 @@
...
+            <p><strong>Reason	:</strong> <%= response.body %></p>
+            <% else %>
+            <h1 font size="24">Your request has been processed successfully</h1>
+            <% end %>
+            
+            <% if @order %>
+            <p><strong>Request details [<%= params[:commit] %>]:</strong></p>
+            <table>
+              <tr><td><strong>Ordered By	:</strong></td>	<td><%= @order[:orderer] %></td></tr>
+              <tr><td><strong>Product	:</strong></td>	<td><%= @order[:product] %></td></tr>
```
bzwei: Why they are not aligned?

bzwei:
@hsong-rh can you take a look at the Hakiri security warning?

app/controllers/api/v1x0/stageaction_controller.rb
```diff
@@ -0,0 +1,85 @@
...
+        case operation
+        when 'Memo', 'Deny'
+          comments = params.require(:message)
+        else
+          comments = params[:message]
+        end
+
+        ManageIQ::API::Common::Request.with_request(@stage.request.context.transform_keys(&:to_sym)) do
+          ActsAsTenant.with_tenant(Tenant.find(@stage.tenant_id)) do
+            ActionCreateService.new(@stage.id).create(
```
bzwei: Right now `ActionCreateService` may raise general `ApprovalError` if the state transition is now allowed. We may need to raise a specific error like `InvalidStateTransitionError`. Here we should catch this error and report a proper error message like `Your action cannot be executed. Somebody else may have approve/denied this request at the mean time.`

app/controllers/api/v1x0/stageaction_controller.rb
```diff
@@ -0,0 +1,92 @@
...
+        render :status => :internal_server_error, :action => :result
+      end
+
+      rescue_from ActionController::ParameterMissing do |e|
+        response.body = e.message
+        render :status => :unprocessable_entity, :action => :result
+      end
+
+      rescue_from Exceptions::InvalidStateTransitionError do |_e|
+        response.body = "Your action cannot be executed. Somebody else may have approve/denied this request at the mean time."
```
bzwei: Typo `approved`

PR #88, opened 2019-04-08
Change when context is stored on Request

lindgrenj6:
https://projects.engineering.redhat.com/browse/SSP-270

Change the way context is stored, so that it only gets set once (when the `Request` is created).
I'll work on changing the MIQ-api-common gem to work with either type of keys as well, that way we don't have to transform the keys in order to use it.


lindgrenj6:
@miq-bot add_reviewer bzwei

hsong-rh:
I did almost the same thing in my PR https://github.com/ManageIQ/approval-api/pull/86.

PR #91, opened 2019-04-11
Restore request context in a separate thread

bzwei:
When creating a new request in another thread we need to pass the current header/tenant to the thread. This is done through reconstructing from the context attribute in the request model.

Fixes https://github.com/ManageIQ/approval-api/issues/83

### DIFF
--- a/app/models/request.rb
+++ b/app/models/request.rb
@@ -21,11 +21,7 @@ class Request < ApplicationRecord
   before_create :set_context
 
   def switch_context
-    if context.present?
-      ManageIQ::API::Common::Request.with_request(context.transform_keys(&:to_sym)) do
-        ActsAsTenant.with_tenant(tenant) { yield }
-      end
-    else
+    ManageIQ::API::Common::Request.with_request(context.transform_keys(&:to_sym)) do
       ActsAsTenant.with_tenant(tenant) { yield }
     end
   end
--- a/app/models/request.rb
+++ b/app/models/request.rb
@@ -20,6 +20,16 @@ class Request < ApplicationRecord
 
   before_create :set_context
 
+  def switch_context
+    if context.present?
+      ManageIQ::API::Common::Request.with_request(context.transform_keys(&:to_sym)) do
+        ActsAsTenant.with_tenant(tenant) { yield }
+      end
+    else
+      ActsAsTenant.with_tenant(tenant) { yield }
+    end
+  end
+
   private
 
   def set_context
--- a/spec/factories/requests.rb
+++ b/spec/factories/requests.rb
@@ -13,12 +13,7 @@ FactoryBot.define do
 
     trait :with_context do
       after(:create) do |obj|
-        obj.update_attributes(:context =>
-          {
-            "headers"      => {"x-rh-identity" => RequestSpecHelper.encoded_user_hash},
-            "original_url" => "http://localhost:3000/api/v1.0/workflows/1/requests"
-          }
-        )
+        obj.update_attributes(:context => RequestSpecHelper.default_request_hash)
       end
     end
 
--- a/spec/factories/requests.rb
+++ b/spec/factories/requests.rb
@@ -10,5 +10,20 @@ FactoryBot.define do
     decision     { :undecided }
 
     workflow
+
+    trait :with_context do
+      after(:create) do |obj|
+        obj.update_attributes(:context =>
+          {
+            "headers"      => {"x-rh-identity" => RequestSpecHelper.encoded_user_hash},
+            "original_url" => "http://localhost:3000/api/v1.0/workflows/1/requests"
+          }
+        )
+      end
+    end
+
+    trait :with_tenant do
+      tenant
+    end
   end
 end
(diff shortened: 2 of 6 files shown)


spec/factories/requests.rb
```diff
@@ -10,5 +10,20 @@
     decision     { :undecided }
 
     workflow
+
+    trait :with_context do
+      after(:create) do |obj|
+        obj.update_attributes(:context =>
+          {
+            "headers"      => {"x-rh-identity"=> "eyJpZGVudGl0eSI6eyJhY2NvdW50X251bWJlciI6IjAzNjkyMzMiLCJ0eXBlIjoiVXNlciIsInVzZXIiOnsidXNlcm5hbWUiOiJqZG9lIiwiZW1haWwiOiJqZG9lQGFjbWUuY29tIiwiZmlyc3RfbmFtZSI6IkpvaG4iLCJsYXN0X25hbWUiOiJEb2UiLCJpc19hY3RpdmUiOnRydWUsImlzX29yZ19hZG1pbiI6ZmFsc2UsImlzX2ludGVybmFsIjpmYWxzZSwibG9jYWxlIjoiZW5fVVMifSwiaW50ZXJuYWwiOnsib3JnX2lkIjoiMzM0MDg1MSIsImF1dGhfdHlwZSI6ImJhc2ljLWF1dGgiLCJhdXRoX3RpbWUiOjYzMDB9fX0="},
```
hsong-rh: You can use `encoded_user_hash`, which is defined in `spec/support/request_spec_helper.rb` as `x-rh-identity`:
```
"headers"      => {"x-rh-identity"=>encoded_user_hash},
```

app/models/request.rb
```diff
@@ -20,6 +20,16 @@ class Request < ApplicationRecord
 
   before_create :set_context
 
+  def switch_context
+    if context.present?
+      ManageIQ::API::Common::Request.with_request(context.transform_keys(&:to_sym)) do
+        ActsAsTenant.with_tenant(tenant) { yield }
+      end
+    else
+      ActsAsTenant.with_tenant(tenant) { yield }
```
mkanoor: @bzwei We should never be without a context, that is an error case and we should raise an exception so that we can trap all these code paths early on.

spec/factories/requests.rb
```diff
@@ -10,5 +10,20 @@
     decision     { :undecided }
 
     workflow
+
+    trait :with_context do
+      after(:create) do |obj|
+        obj.update_attributes(:context =>
+          {
```
mkanoor: @bzwei 
We have a spec helper for this
https://github.com/ManageIQ/catalog-api/blob/148f26523bbd46706eccdeb6ba35c4c90a90d8f9/spec/support/service_spec_helper.rb#L17

I have done more cleanup in this PR for catalog
https://github.com/ManageIQ/catalog-api/pull/228

spec/models/request_spec.rb
```diff
@@ -4,4 +4,25 @@
 
   it { should validate_presence_of(:name) }
   it { should validate_presence_of(:content) }
+
+  describe '#switch_context' do
```
mkanoor: @bzwei Not sure what this is testing. This logic has been tested in other places in the common gem for the request headers. Shouldn't your test case be running the Auto Approval test and checking if it ran in the proper context as opposed to testing the switch_context.

spec/models/request_spec.rb
```diff
@@ -4,4 +4,25 @@
...
+      it 'sets the http request header and current tenant' do
+        request = FactoryBot.create(:request, :with_context)
+        request.switch_context do
+          expect(ManageIQ::API::Common::Request.current.to_h).to eq(request.context.transform_keys(&:to_sym))
+          expect(ActsAsTenant.current_tenant).to eq(request.tenant)
+        end
+      end
+    end
+
+    context 'without context' do
```
mkanoor: @bzwei this test should be testing if exceptions be raised if the context doesn't exist. We will always have context.

spec/factories/requests.rb
bzwei: In your example spec helper why does the `original_url` appear twice?

mkanoor: The original_url is supposed to be part of the "Request" and not the headers. Initially I had added it to the headers and we cleaned it up so that its only part of the request.

spec/models/request_spec.rb
bzwei: This is the unit test for the `#switch_context` method. The test validates current request and tenant are the same as what are recorded in the request context.

app/models/request.rb
```diff
@@ -20,6 +20,12 @@ class Request < ApplicationRecord
 
   before_create :set_context
 
+  def switch_context
+    ManageIQ::API::Common::Request.with_request(context.transform_keys(&:to_sym)) do
+      ActsAsTenant.with_tenant(tenant) { yield }
```
mkanoor: @bzwei This seems like its not in the right place, I had talked to @Fryguy to move this into the common gem so all paths be they come from the controller or background jobs should be able to set this

bzwei: This line is still needed. The common gem does not automatically switch tenant.

PR #92, opened 2019-04-11
Remove context column from all responses via OpenApi::Serializer

lindgrenj6:
Modify the `OpenApi::Serializer` class to remove `context` from the responses anywhere it is set. This is primarily in the `Request` class for now, but can be expanded in the future if needed

### DIFF
--- a/app/controllers/concerns/response.rb
+++ b/app/controllers/concerns/response.rb
@@ -1,5 +1,5 @@
 module Response
   def json_response(object, status = :ok)
-    render json: object, status: status
+    render :json => object.as_json(:except => :context), :status => status
   end
 end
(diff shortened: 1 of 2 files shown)


lindgrenj6:
@miq-bot add_reviewer bzwei

lib/open_api/serializer.rb
```diff
@@ -8,6 +8,8 @@ def as_json(arg = {})
           hash[attr_key] = hash[attr_key].iso8601 if hash[attr_key].kind_of?(Time)
           hash[attr_key] = hash[attr_key].to_s if attr_key.ends_with?("_id") || attr_key == "id"
         end
+
+        hash.delete("context")
```
hsong-rh: I feel it's better to use `:except` options of as_json to exclude `context`, like:
```
as_json(except: :context)
```

lindgrenj6: Ok, updated to use `except:` syntax in the `Response` concern

hsong-rh:
LGTM 👍

PR #93, opened 2019-04-11
Hardcode seeding BPM parameters

bzwei:
In #84 the BPM parameters are passed in through ENV. From the review comments of the PR to create such ENVs in the OCP deployment it is recommended to have these parameters hardcoded in the app. It makes sense because these parameters are for seeding and should not change.


bzwei:
@carbonin please review

PR #98, opened 2019-04-18
Fixed return code to 404

hsong-rh:
https://projects.engineering.redhat.com/browse/SSP-280


hsong-rh:
@bzwei Please review.

PR #101, opened 2019-04-18
Act as org_admin when access rbac groups

bzwei:


app/services/context_service.rb
```diff
@@ -0,0 +1,31 @@
+class ContextService
+  attr_accessor :context
```
syncrou: Should this be an `attr_reader`?

app/services/context_service.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  def initialize(context)
+    self.context = context
```
syncrou: If using an `attr_reader` change to `@context = context`.

app/services/context_service.rb
```diff
@@ -0,0 +1,31 @@
+class ContextService
```
mkanoor: @bzwei Should this class itself be called AdminContext.

mkanoor: @bzwei What would happen if the RBAC team gives us a well defined role that allows us to access the principals in a tenant, how much of the above code would have to change. We need to note that what we are doing is a hack and not an ideal solution for us. So if we have to throw this away and use a real solution we dont have to redo the whole thing all over.

bzwei: I moved `Request#switch_context` to this class. This class will provide all context switching/restoration operations, not limit to admin. We can continue to use this service class to handle changes when RBAC is improved. Anyway we always need to switch a role/context to access groups/principals because they definitely will not be accessible to a regular user.

app/services/context_service.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  def as_org_admin(&block)
+    switch_context(org_admin_context, &block)
+  end
+
+  private
+
+  def org_admin_context
+    decoded = Base64.urlsafe_decode64(context["headers"]["x-rh-identity"])
+    decoded.sub!(/"is_org_admin":false/, '"is_org_admin":true')
```
hsong-rh: In case we may need to change other attributes such as names to `sysadmin`, may we pass this substitute string (`is_org_admin`) as default parameters?

bzwei: This is consider a hack which might be a temporary solution. We don't need to generalize it.

PR #102, opened 2019-04-22
Prevent default workflow from being destroyed

bzwei:
The default workflow that was created from seeding cannot be deleted.


app/models/workflow.rb
```diff
@@ -6,6 +6,8 @@ class Workflow < ApplicationRecord
 
   validates :name, :presence => :name
 
+  before_destroy :not_to_delete_default
```
gmcculloug: Overall, this looks good but suggest we use a different method name.  Maybe `protect_default`

PR #103, opened 2019-04-22
Added uniqueness validation for workflow name

hsong-rh:


app/models/workflow.rb
```diff
@@ -5,6 +5,7 @@ class Workflow < ApplicationRecord
   has_many :requests, -> { order(:id => :asc) }, :inverse_of => :workflow
 
   validates :name, :presence => :name
+  validates :name, :uniqueness => { :scope => :tenant_id }
```
bzwei: should combine two lines into one:
```
validates :name, :presence => true, :uniqueness => { :scope => :tenant_id }
```
However this scope does not prevent a workflow in a tenant from having the same name as another one without a tenant.

gmcculloug: This was my question yesterday.  Is the `scope` needed at all?  I think we should be comparing to the workflows that are returned which should already be scoped to the tenant + global.

Put a different way, you are only creating workflows within a tenant, with the exception of the globally seeded workflows which are done by the system.

cc @syncrou

bzwei: @gmcculloug I tested. `ActsAsTenant` does not automatically add the scope.

gmcculloug: Correct, the controller would set the tenant.  My point was that you really are never creating workflows outside of a tenant which would have applied the scope.  (Again, with the exception of global system workflows.)

So if you are testing you need to use the `ActsAsTenant.with_tenant` as part of the test.  In the catalog repo we updated our tests to account for this.

bzwei: We ship a workflow `Always approve` which belongs to no tenant. We need to prevent an admin from creating another `Always approve` workflow. 
We may create a [custom validation method](https://guides.rubyonrails.org/active_record_validations.html#custom-methods)

gmcculloug: Agreed.  After testing locally I can verify that the validation does not honor the current tenant setting and requires the validation to include the proper tenant scope.  👍

app/models/workflow.rb
```diff
@@ -22,6 +23,12 @@ def self.default_workflow_query
   end
   private_class_method :default_workflow_query
 
+  def unique_with_same_or_no_tenant
+    if Workflow.find_by(:name => name, :tenant_id => tenant_id) || name == "Always approve"
```
bzwei: as simple as ` if Workflow.exists?(:name => name)`

app/models/workflow.rb
```diff
@@ -4,7 +4,8 @@ class Workflow < ApplicationRecord
   belongs_to :template
   has_many :requests, -> { order(:id => :asc) }, :inverse_of => :workflow
 
-  validates :name, :presence => :name
+  validates :name, :presence => true
+  validate :unique_with_same_or_no_tenant, :on => :create
```
bzwei: can we update name? if yes, then remove `:on => :create`

spec/requests/v1.0/requests_spec.rb
```diff
@@ -7,7 +7,7 @@
   let(:tenant) { create(:tenant, :external_tenant => 369_233) }
 
   let!(:template) { create(:template) }
-  let!(:workflow) { create(:workflow, :name => 'Always approve') } #:template_id => template.id) }
+  let!(:workflow) { create(:workflow, :name => 'Test always approve') } #:template_id => template.id) }
```
bzwei: Without seeding why do we need to avoid `Always approve`?

spec/models/workflow_spec.rb
```diff
@@ -11,4 +14,52 @@
...
+        end
+      end
+    end
+  end
+
+  context "with one tenant" do
+    describe "create workflow" do
+      before do
+        Workflow.seed
+        ActsAsTenant.with_tenant(tenant) { workflow }
```
bzwei: You should not need this line.

We can do the test without seeding. Simply do
```
  before { workflow } # create a workflow without tenant
  ActsAsTenant.with_tenant(tenant) do
    expect { Workflow.create!(:name => workflow.name) }.to raise_error
```

spec/models/workflow_spec.rb
```diff
@@ -11,4 +14,52 @@
...
+
+  context "with same name in same tenant" do
+    describe "create workflow" do
+      before do
+        ActsAsTenant.with_tenant(tenant) { workflow }
+      end
+
+      it "create a workflow with same name" do
+        ActsAsTenant.with_tenant(tenant) do
+          expect { Workflow.create!(:name => "same") }.to raise_error(ActiveRecord::RecordInvalid, 'Validation failed: Name has already been taken')
```
bzwei: `expect { Workflow.create!(:name => workflow.name) }.to raise_error`

spec/models/workflow_spec.rb
```diff
@@ -11,4 +14,52 @@
...
+    end
+
+    describe "create workflow" do
+      before do
+        ActsAsTenant.with_tenant(another_tenant) { workflow_tenant }
+        ActsAsTenant.with_tenant(tenant) { workflow }
+      end
+
+      it "return created workflows" do
+        expect(workflow.name).to eq("same")
```
bzwei:
```
expect(workflow_tenant.name).to eq(workflow.name)
```
Suggest to rename `workflow_tenant` to `another_workflow`.

spec/models/workflow_spec.rb
```diff
@@ -11,4 +14,52 @@
       expect(described_class.first.template).to be_nil
     end
   end
+
+  context "with same name in different tenants" do
+    let(:another_tenant) { create(:tenant) }
+    let(:workflow_tenant) do 
+      create(:workflow, :name => "same")
```
bzwei: `:name => workflow.name`

spec/models/workflow_spec.rb
```diff
@@ -11,4 +14,52 @@
...
+    end
+
+    describe "create workflow" do
+      before do
+        ActsAsTenant.with_tenant(another_tenant) { workflow_tenant }
```
bzwei: switch the order of line 26 and 27

spec/models/workflow_spec.rb
```diff
@@ -1,4 +1,7 @@
-RSpec.describe Workflow, type: :model do
+RSpec.describe Workflow, :type => :model do
+  let(:tenant) { create(:tenant) }
+  let(:workflow) { create(:workflow, :name => "same") }
```
bzwei: with the comments below, we don't need to hardcode a name here.

spec/models/workflow_spec.rb
```diff
@@ -11,4 +14,48 @@
...
+
+      it "create a workflow with same name" do
+        ActsAsTenant.with_tenant(tenant) do
+          expect { Workflow.create!(:name => workflow.name) }.to raise_error(ActiveRecord::RecordInvalid, 'Validation failed: Name has already been taken')
+        end
+      end
+    end
+  end
+
+  context "with one tenant" do
```
bzwei: `context "with same name in a tenant and in no tenant"`

PR #104, opened 2019-04-22
Add index filtering via MIQ Common Gem

lindgrenj6:
https://projects.engineering.redhat.com/browse/SSP-307

Adds filtering on the index methods via ManageIQ::API::Common::Filter.

- [x] fix tests
- [x] add docs https://github.com/ManageIQ/manageiq-api-common/pull/57

### DIFF
--- a/public/approval/v1.0/openapi.json
+++ b/public/approval/v1.0/openapi.json
@@ -201,6 +201,9 @@
                     },
                     {
                         "$ref": "#/components/parameters/offset"
+                    },
+                    {
+                        "$ref": "#/components/parameters/filter"
                     }
                 ],
                 "responses": {
@@ -257,6 +260,9 @@
                     },
                     {
                         "$ref": "#/components/parameters/offset"
+                    },
+                    {
+                        "$ref": "#/components/parameters/filter"
                     }
                 ],
                 "responses": {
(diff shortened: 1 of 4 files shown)


lindgrenj6:
@miq-bot add_reviewer bzwei 

public/approval/v1.0/openapi.json
```diff
@@ -1050,27 +1076,27 @@
...
-                                "type": "string",
-                                "format": "date-time",
-                                "description": "Timestamp of creation"
-                            },
-                            "stage_id": {
-                                "type": "string",
-                                "description": "Associated stage id"
-                            }
-                        }
+                        "type": "object"
```
hsong-rh: @lindgrenj6 what's the reason you made these changes? reformat?

lindgrenj6: Yes, when defining the `Out` objects initially it looks like we used `allOf` which includes fields, and then after that declared the extra fields. That _can_ go in the same block as `allOf` but that makes it look like the the object has no fields. So all I did was move the extra field declarations outside of the `allOf` part, so they show up the way the `Api::Doc` parser is expecting.

hsong-rh: 👍

PR #106, opened 2019-04-24
Enabled to list requests by approver's username

hsong-rh:
Enable to query requests based on approver, ex. `GET /requests?approver=jode@acme.com`


app/controllers/api/v1x0/requests_controller.rb
```diff
@@ -14,11 +14,29 @@ def show
...
+        if params[:approver]
+          return index_by_approver
+        end
+
         Workflow.find(params.require(:workflow_id)) if params[:workflow_id] # to validate the workflow exists
         reqs = Request.includes(:stages).filter(params.slice(:requester, :decision, :state, :workflow_id))
         collection(reqs)
       end
 
+      def index_by_approver
```
bzwei: should move to a service class

app/controllers/api/v1x0/requests_controller.rb
```diff
@@ -14,11 +14,29 @@ def show
       end
 
       def index
+        if params[:approver]
+          return index_by_approver
+        end
+
         Workflow.find(params.require(:workflow_id)) if params[:workflow_id] # to validate the workflow exists
```
bzwei: Should move to the same service class

app/controllers/api/v1x0/requests_controller.rb
```diff
@@ -14,11 +14,29 @@ def show
       end
 
       def index
+        if params[:approver]
+          return index_by_approver
```
bzwei: why `return` here? shouldn't we support other filtering (decision, state etc) for an approver?
Please consider the filtering together with #104

hsong-rh: New changes are ready, will wait for #104 is merged.

lindgrenj6:
Not sure if work is still being done on this - but you can use multiple filters on the URL e.g.

`GET /requests?filter[decisoin]=approved&filter[approver]=jdoe@acme.com`

would get all of the approved requests by a certain approver. Just looking at your tests it looks like that would be the main use for it.

More docs here: https://github.com/ManageIQ/manageiq-api-common#manageiqapicommonfilter

hsong-rh:
@lindgrenj6 The filter has to be one of AR's attributes, right? Here approver is not `Request's` attribute. That's the reason to create this PR.

lindgrenj6:
@hsong-rh yes that's correct, I didn't realize that approver wasn't an attribute on `Request`. My bad!

PR #107, opened 2019-04-25
Get all process_options under the scope of org_admin

bzwei:
See review comment below

### DIFF
--- a/app/services/jbpm_process_service.rb
+++ b/app/services/jbpm_process_service.rb
@@ -26,14 +26,17 @@ class JbpmProcessService
   end
 
   def process_options
-    groups =
-      ContextService.new(request.context).as_org_admin do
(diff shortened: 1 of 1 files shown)


app/services/jbpm_process_service.rb
```diff
@@ -26,14 +26,17 @@ def template
...
-      'groups'  => groups.as_json,
-      'stages'  => request.stages.as_json }
+    options = nil
+    ContextService.new(request.context).as_org_admin do
+      groups = request.workflow.group_refs.map { |ref| Group.find(ref) }
+
+      options = {
+        'request' => request.as_json,
+        'groups'  => groups.as_json,
+        'stages'  => request.stages.as_json
```
bzwei: Note `request.stages.as_json` needs to access rbac, therefore the whole method should be placed under the scope of `as_org_admin`. It is the reason this PR is created.

PR #109, opened 2019-04-29
Removed platform field from order content

hsong-rh:


hsong-rh:
This PR is related with: https://github.com/ManageIQ/approval-process/pull/12

PR #110, opened 2019-04-29
Enabled to auto fill requester and processed_by attributes

hsong-rh:
Auto filled attributes of requester and processed_by if they are not set.

### DIFF
--- a/app/models/action.rb
+++ b/app/models/action.rb
@@ -8,7 +8,7 @@ class Action < ApplicationRecord
   DENY_OPERATION    = 'deny'.freeze
   OPERATIONS = [NOTIFY_OPERATION, SKIP_OPERATION, MEMO_OPERATION, APPROVE_OPERATION, DENY_OPERATION].freeze
 
-  validates :operation,    :inclusion => { :in => OPERATIONS }
+  validates :operation, :inclusion => { :in => OPERATIONS }
   validates :processed_by, :presence  => true
   belongs_to :stage, :inverse_of => :actions
 end
--- a/app/services/action_create_service.rb
+++ b/app/services/action_create_service.rb
@@ -7,6 +7,13 @@ class ActionCreateService
 
   def create(options)
     stage_options = validate_operation(options)
+
+    options = options.transform_keys(&:to_sym)
+    unless options[:processed_by]
+      requester = ManageIQ::API::Common::Request.current.user
+      options[:processed_by] = requester.username
+    end
+
     Action.create!(options.merge(:stage => stage)).tap do
       if stage_options
         StageUpdateService.new(stage.id).update(stage_options)
--- a/app/services/request_create_service.rb
+++ b/app/services/request_create_service.rb
@@ -17,12 +17,19 @@ class RequestCreateService
       )
     end
 
+    options = options.transform_keys(&:to_sym)
     create_options = options.merge(
       :workflow => workflow,
       :state    => Request::PENDING_STATE,
       :decision => Request::UNDECIDED_STATUS,
       :stages   => stages
     )
+
+    unless options[:requester]
+      requester = ManageIQ::API::Common::Request.current.user
+      create_options[:requester] = "#{requester.first_name} #{requester.last_name}"
+    end
+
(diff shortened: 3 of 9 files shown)


app/services/request_create_service.rb
```diff
@@ -24,6 +24,7 @@ def create(options)
       :stages   => stages
     )
     Request.create!(create_options).tap do |request|
+      request.update_attributes!(:requester => ContextService.new(request.context).identity_fullname) unless request.requester
```
bzwei: remove it. insert before line 26
```
unless options[:requester]
  requester = ManageIQ::API::Common::Request.current.user
  create_options[:requester] = "#{requester.first_name} #{requester.last_name}"
end
```

app/services/context_service.rb
```diff
@@ -13,6 +13,16 @@ def as_org_admin(&block)
     switch_context(org_admin_context, &block)
   end
 
+  def identity_username
```
bzwei: do not need the two new methods

app/services/action_create_service.rb
```diff
@@ -7,7 +7,8 @@ def initialize(stage_id)
 
   def create(options)
     stage_options = validate_operation(options)
-    Action.create!(options.merge(:stage => stage)).tap do
+    Action.create!(options.merge(:stage => stage)).tap do |action|
```
bzwei: apply the same comment for `request_create_service`

app/models/action.rb
```diff
@@ -8,7 +8,6 @@ class Action < ApplicationRecord
   DENY_OPERATION    = 'deny'.freeze
   OPERATIONS = [NOTIFY_OPERATION, SKIP_OPERATION, MEMO_OPERATION, APPROVE_OPERATION, DENY_OPERATION].freeze
 
-  validates :operation,    :inclusion => { :in => OPERATIONS }
-  validates :processed_by, :presence  => true
+  validates :operation, :inclusion => { :in => OPERATIONS }
```
bzwei: no need to remove validates for processed_by

spec/requests/v1.0/workflows_spec.rb
```diff
@@ -8,7 +8,7 @@
 
   let!(:template) { create(:template) }
   let(:template_id) { template.id }
-  let!(:workflows) { create_list(:workflow, 20, :template_id => template.id) }
+  let!(:workflows) { create_list(:workflow, 16, :template_id => template.id) }
```
bzwei: why 16 or 20?

hsong-rh: 20 exceeds the test boundary due to the extra seeding one

spec/services/action_create_service_spec.rb
```diff
@@ -71,6 +77,13 @@
     end
   end
 
+  context 'auto set processed_by if nil' do
+    it 'create a new action with nil processed_by' do
+      action = svc1.create('operation' => Action::MEMO_OPERATION, 'comments' => 'later')
+      expect(action.processed_by).to eq("jdoe")
```
bzwei: should not hardcode `jdoe`, should be `ManageIQ::API::Common::Request.current.user.username`

spec/services/request_create_service_spec.rb
```diff
@@ -12,6 +12,20 @@
     end
   end
 
+  context 'with auto fill requester' do
+    it 'auto fill requester if it is nil' do
+      request = subject.create(:name => 'req1', :content => 'test me')
+      request.reload
+      expect(request.requester).to eq("John Doe")
```
bzwei: same, get the full name from current header

spec/services/request_create_service_spec.rb
```diff
@@ -12,6 +12,20 @@
...
+    end
+
+    it 'skip to auto fill if it is set' do
```
bzwei: `skips auto filling if requester is set`

spec/services/request_create_service_spec.rb
hsong-rh: There is no `fullname` method in in either `header` or `user`

PR #111, opened 2019-04-29
Fix OpenAPI Generator by removing all 'allOf' calls

lindgrenj6:
https://projects.engineering.redhat.com/browse/SSP-332

The openapi generator was having issues with the allOf calls. Spoke with Bill and the best path forward was to just remove them since they aren't really adding anything.


lindgrenj6:
@miq-bot add_reviewer bzwei

PR #113, opened 2019-05-02
Add Entitlements Check

lindgrenj6:
https://projects.engineering.redhat.com/browse/SSP-369

Similar to https://github.com/ManageIQ/catalog-api/pull/271 , just adding a check on the entitlements array in the headers.


lindgrenj6:
@miq-bot add_reviewer bzwei 

app/controllers/application_controller.rb
```diff
@@ -21,4 +23,10 @@ def current_tenant(current_user)
     return tenant if tenant
     raise  Exceptions::NoTenantError
   end
+
+  def check_entitled(entitlement)
+    required_entitlements = %i[hybrid_cloud?]
```
gmcculloug: This makes the most sense for now.  At some point in the future we'll need to open this up when approval supports other services, but there is no current timeframe for that.

spec/support/request_spec_helper.rb
```diff
@@ -12,6 +12,20 @@ def version(ver = 'v1.0')
   end
 
   DEFAULT_USER = {
+    "entitlements" => {
```
syncrou: This `DEFAULT_USER` constant shouldn't be needed if 
```
Dir[ManageIQ::API::Common::Engine.root.join("spec/support/**/*.rb")].each { |f| require f }
```
Is included. The `default_user_hash` methods in `UserHeaderSpecHelper` will cover all the behavior around the default_user_hash

spec/controllers/v1.0/actions_controller_spec.rb
```diff
@@ -1,2 +1,38 @@
...
+    end
+    let(:missing_hash) do
+      missing_hash = default_user_hash
+      missing_hash.delete("entitlements")
+      missing_hash
+    end
+
+    it "fails if the hybrid_cloud entitlement is false" do
+      headers = { 'x-rh-identity' => encoded_user_hash(false_hash), 'x-rh-insights-request-id' => 'gobbledygook' }
+      get "#{api_version}/workflows", :params => { :limit => 5, :offset => 0 }, :headers => headers
```
bzwei: Does not make sense that you test a `workflow` endpoint in the `actions_controller_spec`.
You can move the tests to any of the existing spec under `requests` folder, where you don't need to prepare variables in order to get an endpoint. 
Later we should move all specs from `requests` to `controllers`.

cc @hsong-rh

lindgrenj6: @bzwei I'm not testing workflow specific functionality (any endpoint would work, I just picked workflows), it's testing `actions_controller.rb` functionality, specifically stuff in the `IndexMixin`. If you want me to move it to the workflows spec file I can, but just thought I would explain why I put this here instead of there.

hsong-rh: @lindgrenj6 Please put this temporarily into one of spec file in `spec/requests`. Later I'll move all of them back to `controllers`.

spec/spec_helper.rb
```diff
@@ -96,6 +97,8 @@
   Kernel.srand config.seed
 =end
 
-  config.include RequestSpecHelper, type: :request
   config.include FactoryBot::Syntax::Methods
+  config.include UserHeaderSpecHelper
```
hsong-rh: Why do you need this `UserHeaderSpecHelper`? Like @syncrou said below, one of them should be enough, right?

lindgrenj6: We do the same thing on Catalog - I was just trying to get it to work, but it still isn't. I'll remove them both as they don't do anything it appears.

spec/support/request_spec_helper.rb
lindgrenj6: Yeah I tried to include the UserHeaderSpecHelper just like we do on catalog, but it isn't acting properly so I fell back to just updating the other hash for now.

spec/spec_helper.rb
```diff
@@ -1,4 +1,5 @@
 Dir[Rails.root.join("spec/support/**/*.rb")].each { |f| require f }
+Dir[ManageIQ::API::Common::Engine.root.join("spec/support/**/*.rb")].each { |f| require f }
```
hsong-rh: You don't need this line. `RequestSpecHelper` does the same thing as `UserHeaderSpecHelper` now.

spec/spec_helper.rb
hsong-rh: Same thing here, no need to include `UserHeaderSpecHelper`

spec/spec_helper.rb
lindgrenj6: Removed it for now, until we can debug why the common gem `spec/support` isn't loading properly.

lindgrenj6:
@miq-bot remove_label unmergeable

PR #114, opened 2019-05-03
Bump manageiq-loggers version to pick up request_id changes.

gmcculloug:
Logging will now include thread local request_id when available.

See https://github.com/ManageIQ/manageiq-loggers/pull/7 (Get the request id from the thread local variable and log it)


gmcculloug:
Updated along with Catalog https://github.com/ManageIQ/catalog-api/pull/287

PR #119, opened 2019-05-16
Return 403 when trying to delete default workflow

bzwei:
The backend already prevent the default workflow from being deleted but the API still returns success.

Addressed QE test on https://projects.engineering.redhat.com/browse/SSP-330

### DIFF
--- a/app/controllers/api/v1x0/workflows_controller.rb
+++ b/app/controllers/api/v1x0/workflows_controller.rb
@@ -25,11 +25,15 @@ module Api
       end
 
       def destroy
-        Workflow.find(params.require(:id)).destroy
-
+        workflow = Workflow.find(params.require(:id))
+        workflow.destroy!
         head :no_content
       rescue ActiveRecord::InvalidForeignKey => e
         json_response({ :message => e.message }, :forbidden)
+      rescue ActiveRecord::RecordNotDestroyed
+        raise unless workflow.errors[:base].include?(Workflow::MSG_PROTECTED_RECORD)
+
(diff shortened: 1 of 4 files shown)


app/controllers/api/v1x0/workflows_controller.rb
```diff
@@ -25,11 +25,17 @@ def index
...
-        json_response({ :message => e.message }, :forbidden)
+        begin
+          workflow.destroy!
+          head :no_content
+        rescue ActiveRecord::InvalidForeignKey => e
+          json_response({ :message => e.message }, :forbidden)
+        rescue ActiveRecord::RecordNotDestroyed
+          raise unless workflow.errors[:base].include?(Workflow::MSG_PROTECTED_RECORD)
+          json_response({ :message => Workflow::MSG_PROTECTED_RECORD }, :forbidden)
+        end
```
gmcculloug: You do not need to define the `begin` for this.

```ruby
def destroy
  workflow = Workflow.find(params.require(:id))
  workflow.destroy!
  head :no_content

rescue ActiveRecord::InvalidForeignKey => e
rescue ActiveRecord::RecordNotDestroyed
  raise unless workflow.errors[:base].include?(Workflow::MSG_PROTECTED_RECORD)
  json_response({ :message => Workflow::MSG_PROTECTED_RECORD }, :forbidden)
end
```

syncrou:
@bzwei can the rubocop warning be resolved?

bzwei:
@syncrou warning is gone

PR #121, opened 2019-05-20
Compliance with IPP-12

mkanoor:
Openapi.json should be fetchable with and without authentication
Incorporate changes made in PR https://github.com/ManageIQ/manageiq-api-common/pull/51

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -8,12 +8,16 @@ class ApplicationController < ActionController::API
 
   def with_current_request
     ManageIQ::API::Common::Request.with_request(request) do |current|
-      raise ManageIQ::API::Common::EntitlementError, "User not Entitled" unless check_entitled(current.entitlement)
+      if current.required_auth?
+        raise ManageIQ::API::Common::EntitlementError, "User not Entitled" unless check_entitled(current.entitlement)
 
-      begin
-        ActsAsTenant.with_tenant(current_tenant(current.user)) { yield }
-      rescue Exceptions::NoTenantError
-        json_response({ :message => 'Unauthorized' }, :unauthorized)
+        begin
+          ActsAsTenant.with_tenant(current_tenant(current.user)) { yield }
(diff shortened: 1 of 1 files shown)


app/controllers/application_controller.rb
```diff
@@ -8,12 +8,16 @@ class ApplicationController < ActionController::API
...
-        ActsAsTenant.with_tenant(current_tenant(current.user)) { yield }
-      rescue Exceptions::NoTenantError
-        json_response({ :message => 'Unauthorized' }, :unauthorized)
+        begin
+          ActsAsTenant.with_tenant(current_tenant(current.user)) { yield }
+        rescue Exceptions::NoTenantError
+          json_response({ :message => 'Unauthorized' }, :unauthorized)
+        end
+      else
+        ActsAsTenant.without_tenant { yield }
```
bzwei: Do we really want to strip off tenant? Based on the api-common, `required_auth?` is the opposite of `optional_auth?`. To play safely I would say let's continue with the tenant if it is given.

mkanoor: @bzwei In all of our use cases we will always have the tenant, the only time it won't be there is when some one comes asking for the openapi.json without credentials. The only use case where there is no tenant is for some inventory refresh other than that all use cases will always have a request and a tenant.

bzwei: I am thinking about future. Currently we accept optional tenant only for `openapi.json`, but the list could expand to other resources, which may require us to act differently based on whether a tenant is present. One analogy is a checking out a shopping cart. A guest cannot receive any reward points while a logged in user gets them.

gmcculloug:
@bzwei @mkanoor Is this PR still required?

app/controllers/application_controller.rb
mkanoor: @gmcculloug In talking with Jason we need to have this shared logic in the common gem so its consistent across services and we dont have services implementing their own flavor.

PR #122, opened 2019-05-21
Add filters validation

hsong-rh:
https://projects.engineering.redhat.com/browse/SSP-315

### DIFF
--- a/app/controllers/api/v1x0/requests_controller.rb
+++ b/app/controllers/api/v1x0/requests_controller.rb
@@ -14,6 +14,8 @@ module Api
       end
 
       def index
+        validate_filters
+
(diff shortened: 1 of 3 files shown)


bzwei:
I feel we may just want to update the api spec rather than validate in the code. Return an empty list and code 200 seem fine to me.

bzwei:
@hsong-rh per our discussion, please clean up `openopi.json` that we don't specifically list allowed filtered attributes for requests because the filtering is handled in a common way for listing any resources. As a result returning code 200 with an empty array is valid.

hsong-rh:
@bzwei Cleanup on `openapi.json` is done, please review.

PR #123, opened 2019-05-24
Allow to set description of request

bzwei:
Addresses the issue that the description field of the request was not saved.
Fixes https://projects.engineering.redhat.com/browse/SSP-454

### DIFF
--- a/app/controllers/api/v1x0/requests_controller.rb
+++ b/app/controllers/api/v1x0/requests_controller.rb
@@ -28,7 +28,7 @@ module Api
       private
 
       def request_params
-        params.permit(:name, :requester, :content => {})
+        params.permit(:name, :description, :requester, :content => {})
       end
     end
   end
--- a/spec/controllers/v1.0/requests_controller_spec.rb
+++ b/spec/controllers/v1.0/requests_controller_spec.rb
@@ -81,7 +81,7 @@ RSpec.describe Api::V1x0::RequestsController, :type => :request do
   end
 
   # Test suite for GET /requests?state=
-  describe 'GET /requests?state=notified' do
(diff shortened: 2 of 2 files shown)


hsong-rh:
One more place for the filter needs to be changed:

In `public/approval/v1.0/openapi.json,` all query parameters about `state/decision/requester` on `requests` have to be removed:

https://github.com/ManageIQ/approval-api/blob/master/public/approval/v1.0/openapi.json#L190 to L198;
https://github.com/ManageIQ/approval-api/blob/master/public/approval/v1.0/openapi.json#L987 to L1030

Do you want to add changes here or to open another PR for this?

spec/controllers/v1.0/requests_controller_spec.rb
```diff
@@ -81,7 +81,7 @@
   end
 
   # Test suite for GET /requests?state=
-  describe 'GET /requests?state=notified' do
+  describe 'GET /requests?filter[state]=notified' do
```
gmcculloug: In the future these unrelated changes would preferable be moved to a separate PR or at least a separate commit as it is unrelated to the other work in this PR.

app/controllers/api/v1x0/requests_controller.rb
```diff
@@ -28,7 +28,7 @@ def index
       private
 
       def request_params
-        params.permit(:name, :requester, :content => {})
+        params.permit(:name, :description, :requester, :content => {})
```
gmcculloug: @bzwei Are any changes required to `openapi.json`  for this?

bzwei: No. The bug was open because the spec says description is optional, but code ignored it.

PR #127, opened 2019-06-07
Backend support of cancel action

bzwei:
Adding new action type `cancel`.
When cancel action is added to a stage, it cancels the stage and eventually cancels the request.
When a request is canceled, it sends `request_canceled` event to the message bus.

@hsong-rh please add new endpoint `POST /requests/<id>/actions`. We will discuss wither to retire `POST /stages/<id>/actions`

### DIFF
--- a/spec/services/action_create_service_spec.rb
+++ b/spec/services/action_create_service_spec.rb
@@ -48,7 +48,7 @@ RSpec.describe ActionCreateService do
       stage2.reload
       request.reload
       expect(action).to  have_attributes(:operation => Action::DENY_OPERATION, :processed_by => 'man', :comments => 'bad')
-      expect(stage2).to  have_attributes(:state => Stage::FINISHED_STATE,   :decision => Stage::DENIED_STATUS, :reason => 'bad')
+      expect(stage2).to  have_attributes(:state => Stage::FINISHED_STATE,   :decision => Stage::DENIED_STATUS,   :reason => 'bad')
       expect(request).to have_attributes(:state => Request::FINISHED_STATE, :decision => Request::DENIED_STATUS, :reason => 'bad')
     end
   end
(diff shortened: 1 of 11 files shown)


bzwei:
@hsong-rh the business-process side should skip the rest stages after current stage is canceled.

spec/services/action_create_service_spec.rb
```diff
@@ -60,9 +61,23 @@
...
+  context 'cancel operation' do
+    it 'updates stage and request' do
+      stage1.update_attributes(:state => Stage::NOTIFIED_STATE)
+      action1 = svc1.create('operation' => Action::CANCEL_OPERATION, 'processed_by' => 'requester', 'comments' => 'regret')
+      stage1.reload
+      stage2.reload
+      request.reload
+      expect(action1).to have_attributes(:operation => Action::CANCEL_OPERATION, :processed_by => 'requester', :comments => 'regret')
+      expect(stage1).to  have_attributes(:state => Stage::CANCELED_STATE,   :decision => Stage::UNDECIDED_STATUS, :reason => 'regret')
+      expect(stage2).to  have_attributes(:state => Stage::SKIPPED_STATE,    :decision => Stage::UNDECIDED_STATUS, :reason => nil)
```
syncrou: I'm wondering if `stage2` should also be set to `CANCELED_STATE` as opposed to `SKIPPED_STATE`? Same with the `decision` as `UNDECIDED_STATUS` for each.  It seems like all the stages should have the cancelled state and the reason set to a 'CANCELED_STATUS'.  

In my mind though, I'm seeing a 'cancel` as a definite action with a reason of 'canceled'.  Possibly that is not how the backend behavior models this decision.

bzwei: The backend implementation has a distinguish of system role vs human role. `skip` action and its corresponding `skipped` state are done by system (PAM in this case), while other actions such as `approve`, `deny`, and `cancel` are done by human. If a certain stage is `denied` by human then the system decides to `skip` the rest stages. Here we are following the same pattern for canceling. No code change is needed, i.e., the system always skips unneeded stages, not cancel.

As for the status, I agree we may be able to introduce a `CANCELED_STATUS`.

eclarizio:
@bzwei @hsong-rh Not sure if it should have been done here or in #128, but I think we need to add "cancel" to the enum list in the openapi.json file for the `ActionIn` schema reference, right?

hsong-rh:
@eclarizio I'll create a separate PR to modify `ActionIn` in openapi.json.

PR #128, opened 2019-06-11
Added requests/:id/actions endpoint

hsong-rh:
Add an endpoint to cancel the request: requests/:id/cancel

This PR is related with: https://github.com/ManageIQ/approval-api/pull/127

### DIFF
--- a/app/controllers/api/v1x0/requests_controller.rb
+++ b/app/controllers/api/v1x0/requests_controller.rb
@@ -4,17 +4,8 @@ module Api
       include Mixins::IndexMixin
 
       def create
-        if params[:request_id]
-          req = Request.find(params.require(:request_id))
-          current_stage = req.current_stage
-          raise Exceptions::ApprovalError, "Request has finished its lifecycle. No more action can be added to its current stage." unless current_stage
-
-          action = ActionCreateService.new(current_stage.id).create(action_params)
-          json_response(action, :created)
-        else
-          req = RequestCreateService.new(params.require(:workflow_id)).create(request_params)
-          json_response(req, :created)
-        end
+        req = RequestCreateService.new(params.require(:workflow_id)).create(request_params)
+        json_response(req, :created)
       end
 
       def show
@@ -38,10 +29,6 @@ module Api
 
       def request_params
         params.permit(:name, :description, :requester, :content => {})
-      end
-
-      def action_params
-        params.permit(:operation, :processed_by, :comments)
       end
     end
   end
--- a/app/controllers/api/v1x0/requests_controller.rb
+++ b/app/controllers/api/v1x0/requests_controller.rb
@@ -4,8 +4,17 @@ module Api
       include Mixins::IndexMixin
 
       def create
-        req = RequestCreateService.new(params.require(:workflow_id)).create(request_params)
-        json_response(req, :created)
+        if params[:request_id]
+          req = Request.find(params.require(:request_id))
+          current_stage = req.current_stage
+          raise Exceptions::ApprovalError, "Request has finished its lifecycle. No more action can be added to its current stage." unless current_stage
+
+          action = ActionCreateService.new(current_stage.id).create(action_params)
+          json_response(action, :created)
+        else
+          req = RequestCreateService.new(params.require(:workflow_id)).create(request_params)
+          json_response(req, :created)
+        end
       end
 
       def show
@@ -25,23 +34,14 @@ module Api
         collection(reqs)
       end
 
-      def cancel
-        req = Request.find(params.require(:request_id))
-        raise Exceptions::ApprovalError, "Unable to cancel request." unless req.current_stage
-
-        ActionCreateService.new(req.current_stage.id).create(
-          :operation    => Action::CANCEL_OPERATION,
-          :processed_by => req.requester,
-          :comments     => params[:comments]
-        )
-
-        head :no_content
-      end
-
       private
 
       def request_params
         params.permit(:name, :description, :requester, :content => {})
+      end
+
+      def action_params
+        params.permit(:operation, :processed_by, :comments)
       end
     end
   end
--- a/app/controllers/api/v1x0/requests_controller.rb
+++ b/app/controllers/api/v1x0/requests_controller.rb
@@ -25,6 +25,19 @@ module Api
         collection(reqs)
       end
 
+      def cancel
+        req = Request.find(params.require(:request_id))
+        raise Exceptions::ApprovalError, "Unable to cancel request." unless req.current_stage
+
+        ActionCreateService.new(req.current_stage.id).create(
+          :operation    => Action::CANCEL_OPERATION,
+          :processed_by => req.requester,
+          :comments     => params[:comments]
+        )
+
+        head :no_content
+      end
+
       private
 
       def request_params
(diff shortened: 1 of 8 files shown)


hsong-rh:
@bzwei @syncrou Please review.

app/controllers/api/v1x0/requests_controller.rb
```diff
@@ -25,6 +25,19 @@ def index
         collection(reqs)
       end
 
+      def cancel
+        req = Request.find(params.require(:request_id))
+        raise Exceptions::ApprovalError, "Unable to cancel request." unless req.current_stage
```
eclarizio: This naming seems odd. It reads like "Do this unless the current stage is....?" Perhaps `req.past_cancellation?` or `req.processing?` or something similar would be better.

hsong-rh: `current_stage` really means the current processing stage of the request. It will be used later to create action on it. Request is finished when `current_stage is nil`.

eclarizio: Yeah I understand what it means, but it just reads strangely because you would expect that there are some stages that are unable to be canceled. Right now it looks like as long as there is a current stage, it's cancelable (which might be true, but again, it just reads strangely).

Also, since you are using it below in the action create, you should probably localize it in a variable so you're not hitting the database both times you call `.current_stage`.

bzwei:
@eclarizio @hsong-rh yes, the endpoint should be generic, not limit to cancel only.

app/controllers/api/v1x0/requests_controller.rb
bzwei: We can introduce a method like `cancelable?` in request. Localizing is not important here. It does not hit the database multiple times.

eclarizio: I realize this is probably going to change, but `current_stage` right now calls `find_by`, and then again on line 32 we're calling `req.current_stage.id`. Does this not call `find_by` twice?

hsong-rh: @bzwei After endpoint changes to `requests/:id/actions`, it's only limit to `cancel` only. So `cancelable?` is unsuitable here. How about  `actionable?`

app/controllers/api/v1x0/requests_controller.rb
```diff
@@ -4,8 +4,16 @@ class RequestsController < ApplicationController
       include Mixins::IndexMixin
 
       def create
-        req = RequestCreateService.new(params.require(:workflow_id)).create(request_params)
-        json_response(req, :created)
+        if params[:request_id]
+          req = Request.find(params.require(:request_id))
+          raise Exceptions::ApprovalError, "Request has been processed!." unless req.current_stage
```
bzwei: `Request has finished its lifecycle. No more action can be added to its current stage.`

app/controllers/api/v1x0/requests_controller.rb
```diff
@@ -4,8 +4,16 @@ class RequestsController < ApplicationController
...
+
+          action = ActionCreateService.new(req.current_stage.id).create(action_params)
```
bzwei: Yes. Here you can locally store resolved `req.current_stage`.

app/models/request.rb
```diff
@@ -21,7 +21,11 @@ class Request < ApplicationRecord
   before_create :set_context
 
   def as_json(options = {})
-    super(options.merge(:methods => [:total_stages, :active_stage]))
+    super(options.merge(:methods => [:total_stages, :active_stage_number]))
+  end
+
+  def current_stage
+    stages.find_by(:state => [Stage::NOTIFIED_STATE, Stage::PENDING_STATE])
```
bzwei: `find_by` is a db operation while `find_index` in the other method is an array operation. Can you guarantee `find_by` returns the first sorted qualified stage?

config/routes.rb
```diff
@@ -25,6 +25,7 @@ def put(*_args)
 
       resources :requests, :only => %i(index show) do
         resources :stages, :only => [:index]
+        post :actions, :action => 'create', :controller => 'requests'
```
bzwei: why the syntax is different from workflows/id/requests or templates/id/workflows? shouldn't you use actions controller?

spec/controllers/v1.0/actions_controller_spec.rb
```diff
@@ -59,4 +58,40 @@
       end
     end
   end
+
+  describe 'POST /requests/:request_id/actions' do
```
bzwei: You've added the spec in `actions_controller_spec` indicating it belongs to actions controller.

spec/models/request_spec.rb
```diff
@@ -28,19 +29,21 @@
       end
 
       it 'has active_stage pointing to the first stage' do
-        expect(subject.as_json).to include('active_stage' => 3, 'total_stages' => 3)
+        expect(subject.as_json).to include('active_stage_number' => 3, 'total_stages' => 3)
```
bzwei: previous line has a typo, should be `has active_stage pointing to the last stage`

app/models/request.rb
```diff
@@ -21,7 +21,11 @@ class Request < ApplicationRecord
   before_create :set_context
 
   def as_json(options = {})
-    super(options.merge(:methods => [:total_stages, :active_stage]))
+    super(options.merge(:methods => [:total_stages, :active_stage_number]))
```
bzwei: This breaks the UI. Let's discuss

spec/controllers/v1.0/actions_controller_spec.rb
```diff
@@ -59,4 +58,40 @@
...
+
+      it 'returns status code 201' do
+        expect(req.stages.first.state).to eq(Stage::CANCELED_STATE)
+        expect(req.stages.last.state).to eq(Stage::SKIPPED_STATE)
+        expect(response).to have_http_status(201)
+      end
+    end
+
+    context 'when request is not actionable' do
+      let(:req) { create(:request, :with_context, :tenant_id => tenant.id) }
```
syncrou: Do we need to create `:req` for each context?

spec/controllers/v1.0/actions_controller_spec.rb
```diff
@@ -59,4 +58,40 @@
...
+
+  describe 'POST /requests/:request_id/actions' do
+    context 'when request is actionable' do
+      let(:req) { create(:request, :with_context, :tenant_id => tenant.id) }
+      let!(:stage1) { create(:stage, :id => "1", :state => Stage::NOTIFIED_STATE, :request => req, :tenant_id => tenant.id) }
+      let!(:stage2) { create(:stage, :id => "2", :state => Stage::PENDING_STATE, :request => req, :tenant_id => tenant.id) }
+      let(:valid_attributes) { { :operation => 'cancel', :processed_by => 'abcd' } }
+
+      before do
+        allow(Group).to receive(:find)
```
syncrou: Would this `before` block work at the describe level so it doesn't have to be created at each context?

app/models/request.rb
hsong-rh: Based on Rails doc, `find_by` will guarantee to return the first matching record based on the condition.

spec/controllers/v1.0/actions_controller_spec.rb
hsong-rh: Yes, has to link stages to it in each context.

PR #129, opened 2019-06-12
Modify as_json to only expose what is configured in the OpenAPI Spec

lindgrenj6:
https://projects.engineering.redhat.com/browse/SSP-476

This PR modifies `as_json` to return in the response only what is in the OpenAPI Spec.
Largely an implementation of what we did on Catalog here:
 https://github.com/ManageIQ/catalog-api/pull/298

### DIFF
--- a/lib/open_api/serializer.rb
+++ b/lib/open_api/serializer.rb
@@ -1,14 +1,43 @@
 module OpenApi
   module Serializer
     def as_json(arg = {})
-      return super unless kind_of?(ApplicationRecord)
+      previous = super
+      encrypted_columns_set = (self.class.try(:encrypted_columns) || []).to_set
+      encryption_filtered = previous.except(*encrypted_columns_set)
+      version = version_from_path(arg)
 
-      super.tap do |hash|
-        hash.each_key do |attr_key|
-          hash[attr_key] = hash[attr_key].iso8601 if hash[attr_key].kind_of?(Time)
-          hash[attr_key] = hash[attr_key].to_s if attr_key.ends_with?("_id") || attr_key == "id"
-        end
+      return encryption_filtered unless arg.key?(:prefixes) || version
+
+      # We use "#{klass.name}Out" for the definitions on what are returned.
+      schema = Api::Docs[version].definitions[self.class.name + "Out"]
+
+      filter(encryption_filtered.slice(*schema["properties"].keys), schema, encrypted_columns_set)
+    end
+
+    private
+
+    def version_from_path(arg)
+      url = ManageIQ::API::Common::Request.current.instance_variable_get(:@original_url) || nil
+      base_path = url ? URI.parse(url).path : arg[:prefixes].first
+      api_version_from_prefix(base_path)
+    end
+
+    def filter(attrs, schema, encrypted_columns_set)
+      schema["properties"].keys.each do |name|
+        next if attrs[name].nil?
+
+        attrs[name] = attrs[name].iso8601 if attrs[name].kind_of?(Time)
+        attrs[name] = attrs[name].to_s if name.ends_with?("_id") || name == "id"
+        attrs[name] = public_send(name) if !attrs.key?(name) && !encrypted_columns_set.include?(name)
       end
+      attrs.compact
+    end
+
+    def api_version_from_prefix(prefix)
+      return unless prefix
+
+      /\/?\w+\/v(?<major>\d+)[x\.]?(?<minor>\d+)?\// =~ prefix
+      [major, minor].compact.join(".")
     end
   end
 end
--- a/public/approval/v1.0/openapi.json
+++ b/public/approval/v1.0/openapi.json
@@ -1053,11 +1053,6 @@
                     "id": {
                         "type": "string"
                     },
-                    "created_at": {
-                        "type": "string",
-                        "format": "date-time",
-                        "description": "Timestamp of creation"
-                    },
                     "stage_id": {
                         "type": "string",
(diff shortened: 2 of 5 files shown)


lindgrenj6:
@miq-bot add_reviewer bzwei

lindgrenj6:
cc @syncrou @gmcculloug

public/approval/v1.0/openapi.json
```diff
@@ -1082,6 +1077,16 @@
                         "type": "string",
                         "description": "Comments for action",
                         "readOnly": true
+                    },
+                    "created_at": {
+                        "type": "string",
+                        "format": "date-time",
+                        "description": "Timestamp of creation"
+                    },
+                    "updated_at": {
```
bzwei: Are all these timestamps required by the serializer or do you add them just because you think good to have?

lindgrenj6: A few of them were required by the unit tests, but I could also change those tests to look at different fields as well. I'm open to either option. On the Catalog side we added created_at/updated_at to most responses it looked like, I'll double check to see if it's on every response for sure.

lib/open_api/serializer.rb
```diff
@@ -1,14 +1,43 @@
...
+
+        attrs[name] = attrs[name].iso8601 if attrs[name].kind_of?(Time)
+        attrs[name] = attrs[name].to_s if name.ends_with?("_id") || name == "id"
+        attrs[name] = public_send(name) if !attrs.key?(name) && !encrypted_columns_set.include?(name)
       end
+      attrs.compact
+    end
+
+    def api_version_from_prefix(prefix)
+      return unless prefix
```
bzwei: protecting this line does not prevent method `as_json` from crashing. you can just drop it.

lib/open_api/serializer.rb
```diff
@@ -1,14 +1,43 @@
...
+      schema = Api::Docs[version].definitions[self.class.name + "Out"]
+
+      filter(encryption_filtered.slice(*schema["properties"].keys), schema, encrypted_columns_set)
+    end
+
+    private
+
+    def version_from_path(arg)
+      url = ManageIQ::API::Common::Request.current.instance_variable_get(:@original_url) || nil
+      base_path = url ? URI.parse(url).path : arg[:prefixes].first
```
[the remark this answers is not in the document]
bzwei: Who needs `OpenApi::Serializer#as_json`? If only used by REST call then we can assume `url` is always there, same to version and schema. Error handling is not needed since it is a private method.
If the method is also needed for backend test, for example, from rails c, we need error handling and return a JSON to include all attributes.

lib/open_api/serializer.rb
```diff
@@ -1,14 +1,43 @@
...
 
-      super.tap do |hash|
-        hash.each_key do |attr_key|
-          hash[attr_key] = hash[attr_key].iso8601 if hash[attr_key].kind_of?(Time)
-          hash[attr_key] = hash[attr_key].to_s if attr_key.ends_with?("_id") || attr_key == "id"
-        end
+      return encryption_filtered unless arg.key?(:prefixes) || version
+
+      # We use "#{klass.name}Out" for the definitions on what are returned.
+      schema = Api::Docs[version].definitions[self.class.name + "Out"]
```
bzwei: how does it work if `version.nil? && arg.key?(:prefixes)`?

lib/open_api/serializer.rb
```diff
@@ -1,14 +1,43 @@
 module OpenApi
   module Serializer
     def as_json(arg = {})
-      return super unless kind_of?(ApplicationRecord)
+      previous = super
```
bzwei: Suggest the following sequence
```
def as_json(arg = {})
  previous = super  
  encrypted_columns_set = (self.class.try(:encrypted_columns) || []).to_set
  version = version_from_path(arg)

  if version
    schema = Api::Docs[version].definitions[self.class.name + "Out"]
    # in method filter, do not directly update attrs, start with an empty hash
    filter(previous, schema, encrypted_columns_set)
  else
    previous.except(*encrypted_columns_set)
  end
end
```

public/approval/v1.0/openapi.json
syncrou: @bzwei on catalog we ended up exposing timestamps for most of the model backed endpoints. The thought was that dates were useful in general as part of a default json payload. 

We also were expecting a number of timestamps in the original specs, so that also contributed to the changes in the `openapi.json` spec file.

lib/open_api/serializer.rb
syncrou: If version is nil then everything stops.  We could add a raise to help troubleshoot why `version` is nil, but since it has to do with the request, if version doesn't return than nothing can move forward with the serializer.

lib/open_api/serializer.rb
syncrou: @bzwei to give some context here.  The original serializer code in common is here.  https://github.com/ManageIQ/manageiq-api-common/blob/master/lib/manageiq/api/common/open_api/serializer.rb

I would like to pull all the `prefix` checking out. It appears it was added to help test the serializer more than anything.  I left it in on Catalog to help possibly ease our ability to use the manageiq-api-common version with the least amount of changes.

bzwei:
@lindgrenj6 per our discussion with @syncrou we can hold off this PR. The proper enhancement should be done on common gem.

lindgrenj6:
@bzwei closing this for now as the enhancements will be done in common.

PR #130, opened 2019-06-13
Add Jenkinsfile-smoke.groovy to be used for smoke test phase

zsadeh:
Add Jenkinsfile-smoke.groovy to be used for smoke test phase

### DIFF
--- a/Jenkinsfile-smoke.groovy
+++ b/Jenkinsfile-smoke.groovy
@@ -1,0 +1,16 @@
+/*
+ * Requires: https://github.com/RedHatInsights/insights-pipeline-lib
+ */
+
+@Library("github.com/RedHatInsights/insights-pipeline-lib") _
+
+
+if (env.CHANGE_ID) {
+    runSmokeTest (
+        ocDeployerBuilderPath: "approval",
+        ocDeployerComponentPath: "approval",
+        ocDeployerServiceSets: "approval",
+        iqePlugins: ["iqe-approval-plugin"],
+        pytestMarker: "approval-smoke",
+    )
+}
--- a/Jenkinsfile-smoke.groovy
+++ b/Jenkinsfile-smoke.groovy
@@ -9,8 +9,8 @@ if (env.CHANGE_ID) {
     runSmokeTest (
         ocDeployerBuilderPath: "approval",
         ocDeployerComponentPath: "approval",
-        ocDeployerServiceSets: "approval",
+        ocDeployerServiceSets: "approval, platform-mq",
         iqePlugins: ["iqe-approval-plugin"],
-        pytestMarker: "approval-smoke",
+        pytestMarker: "approval-api-smoke",
     )
 }


zsadeh:
@bzwei @hsong-rh 
Added the jenkins file to be used in smoke testing phase.
Please see https://github.com/RedHatInsights/e2e-deploy/blob/master/docs/pipeline.md

zsadeh:
@bsquizz Hey, can you review the Jenkinsfile-smoke.groovy ? 
I just found out that I need to add  "platform-mq" to  ocDeployerServiceSets to deploy kafka message queue. Is that right?

Jenkinsfile-smoke.groovy
```diff
@@ -0,0 +1,16 @@
...
+ */
+
+@Library("github.com/RedHatInsights/insights-pipeline-lib") _
+
+
+if (env.CHANGE_ID) {
+    runSmokeTest (
+        ocDeployerBuilderPath: "approval",
+        ocDeployerComponentPath: "approval",
+        ocDeployerServiceSets: "approval",
```
bsquizz: Does approval interact with any other service sets that you will want to test?

zsadeh: yes. rbac.

bsquizz: Right, ok. And that's not in e2e-deploy yet... we can work on that this week. So I guess to get things started you can just deploy approval ;)

gmcculloug: @zsadeh What about kafka?  I see the catalog version includes `platform-mq`

See https://github.com/ManageIQ/catalog-api/pull/308/files

cc @bzwei

Jenkinsfile-smoke.groovy
```diff
@@ -0,0 +1,16 @@
...
+        iqePlugins: ["iqe-approval-plugin"],
+        pytestMarker: "approval-smoke",
```
gmcculloug: @bsquizz Is there a recommendation for naming of `pytestMarker`?  The catalog project is using `catalog_api_smoke` which includes `api` and uses underscores.  Would be good to have consistency between the projects.

Ref https://github.com/ManageIQ/catalog-api/pull/308/files

Jenkinsfile-smoke.groovy
bzwei: `approval-api` has dependencies on `kafka` and `approval-process`, but they are optional. It depends on what are included in the smoke tests. If `kafka` is not present errors are logged but the request will not fail. If `approval-process` is not present, you cannot test email notification.

We however need to consider the code promotion process for `approval-process`. Let's do it after this work.

Jenkinsfile-smoke.groovy
bzwei: Yes please change to `approval-api-smoke` because we will have `approval-process-smoke`.

bzwei: Yes please change to `approval-api-smoke` because we will have `approval-process-smoke`.

zsadeh: done. I changed to approval-api-smoke

zsadeh: also added platform-mq to enable kafka

bsquizz: > @bsquizz Is there a recommendation for naming of `pytestMarker`? The catalog project is using `catalog_api_smoke` which includes `api` and uses underscores. Would be good to have consistency between the projects.
> 
> Ref https://github.com/ManageIQ/catalog-api/pull/308/files

It's an arbitrary name -- as long as it is unique within our test framework -- it can be whatever name dev & QE thinks fits best :)

Jenkinsfile-smoke.groovy
bsquizz: At the moment, since `approval` is listed as a service set to deploy, the whole `approval` service set will be deployed, which includes all the components here: https://github.com/RedHatInsights/e2e-deploy/tree/master/templates/approval

bzwei: @bsquizz Thanks for the clarification

gmcculloug:
cc @carbonin @Fryguy

carbonin:
Yeah, :+1: here from me also. Just need @Fryguy to make some repo settings changes for the insights bot I think.

PR #132, opened 2019-06-19
Fixed code climate errors

hsong-rh:
https://github.com/codeclimate/codeclimate-rubocop/issues/166

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -28,4 +28,4 @@ plugins:
   rubocop:
     enabled: true
     config: ".rubocop_cc.yml"
-    channel: "rubocop-0-68"
+    channel: "rubocop-0-69"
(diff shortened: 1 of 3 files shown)


hsong-rh:
Only worked in my local env. Closed.

gmcculloug:
@syncrou Should we do the same here as https://github.com/ManageIQ/catalog-api/pull/305

syncrou:
@hsong-rh Per what @gmcculloug suggested.  Applying the fix in ManageIQ/catalog-api#305 did fix the issue for us in Catalog.

Fryguy:
:-1: ... we should not proliferate what is in ManageIQ/catalog-api#305 as it's not maintainable from an org perspective.  Can we please discuss in the miq-bot channel?

syncrou:
@Fryguy  per this comment, was only temporary.

https://github.com/ManageIQ/catalog-api/pull/305#issuecomment-507395730

PR #134, opened 2019-06-25
Remove Basic_auth from openapi because the backend does not support it

bzwei:
It actually causes trouble in the client used by QE.

### DIFF
--- a/spec/controllers/v1.0/actions_controller_spec.rb
+++ b/spec/controllers/v1.0/actions_controller_spec.rb
@@ -66,8 +66,8 @@ RSpec.describe Api::V1x0::ActionsController, :type => :request do
     end
 
     context 'when request is actionable' do
-      let!(:stage1) { create(:stage, :id => "1", :state => Stage::NOTIFIED_STATE, :request => req, :tenant_id => tenant.id) }
-      let!(:stage2) { create(:stage, :id => "2", :state => Stage::PENDING_STATE, :request => req, :tenant_id => tenant.id) }
+      let!(:stage1) { create(:stage, :state => Stage::NOTIFIED_STATE, :request => req, :tenant_id => tenant.id) }
+      let!(:stage2) { create(:stage, :state => Stage::PENDING_STATE, :request => req, :tenant_id => tenant.id) }
       let(:valid_attributes) { { :operation => 'cancel', :processed_by => 'abcd' } }
 
       it 'returns status code 201' do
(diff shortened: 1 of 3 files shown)


bzwei:
@gmcculloug can you merge it? QE is waiting for it.

mkanoor:
@bzwei How would this work for external facing, customers who are using the RHN login which uses userid/password? That is basic auth inside of HTTPS

mkanoor:
@bzwei At least from the catalog perspective we use the basic auth so that customers using the API externally can access it.
https://github.com/mkanoor/catalog-api-client-ruby/blob/b476fbe6a9695f0206be79cce8eff2f527dd78d9/lib/catalog-api-client/configuration.rb#L192

bsquizz:
Aren't all the apps in the public-facing environments accessed via 3scale, with 3scale handling the auth and passing on an `x-rh-identity` header? Your app is not directly receiving a Basic authorization header is it?

syncrou:
@bsquizz once on any of the environments yes.  But we have been using basic auth when testing one service external to 3scale that requires another service behind 3scale to complete that test.

bsquizz:
Does disabling basic auth in the spec affect how the app works? It will still accept basic auth even if the spec changes, or are you generating your app code from the spec?

bzwei:
@mkanoor @syncrou Our backend does not validate username/password, how does this requirement in openapi help? Zvika is facing difficulties currently when doing smoking tests. I guess Beni will have similar problem too.

syncrou:
@bsquizz we're not generating app code from the spec. I just wanted to make sure that we could still use basic auth whether it existed in the spec or not. Thanks.

spec/controllers/v1.0/actions_controller_spec.rb
```diff
@@ -66,8 +66,8 @@
     end
 
     context 'when request is actionable' do
-      let!(:stage1) { create(:stage, :id => "1", :state => Stage::NOTIFIED_STATE, :request => req, :tenant_id => tenant.id) }
-      let!(:stage2) { create(:stage, :id => "2", :state => Stage::PENDING_STATE, :request => req, :tenant_id => tenant.id) }
```
syncrou: Were these changes needed as part of the basic auth change?

bzwei: no. it failed the build. was introduced by a previous PR.

zsadeh:
@gmcculloug 
Well I guess this PR should be closed as we decided not to remove security settings from spec file.
I guess we need a new ticket to add all supported security scheme to the spec file of approval and catalog.

gmcculloug:
@zsadeh Agree to close.  Please open a git issue for the followup work.  Thanks.

PR #138, opened 2019-06-27
Added OwnerField to Request model

hsong-rh:
This is the first phase to enable RBAC support in Approval service: renamed attribute `requester` to `owner` in active record `Request`, and populate it with user information whenever request is created.

### DIFF
--- a/app/controllers/api/v1x0/requests_controller.rb
+++ b/app/controllers/api/v1x0/requests_controller.rb
@@ -28,7 +28,7 @@ module Api
       private
 
       def request_params
-        params.permit(:name, :description, :requester, :content => {})
+        params.permit(:name, :description, :owner, :content => {})
       end
     end
   end
--- a/app/controllers/api/v1x0/requests_controller.rb
+++ b/app/controllers/api/v1x0/requests_controller.rb
@@ -28,7 +28,7 @@ module Api
       private
 
       def request_params
-        params.permit(:name, :description, :content => {})
+        params.permit(:name, :description, :requester_name, :content => {})
       end
     end
   end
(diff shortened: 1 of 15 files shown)


app/controllers/api/v1x0/requests_controller.rb
```diff
@@ -28,7 +28,7 @@ def index
       private
 
       def request_params
-        params.permit(:name, :description, :requester, :content => {})
+        params.permit(:name, :description, :owner, :content => {})
```
bzwei: We can drop `owner` from the api and remove it from `openapi.json` too. There was a Jira ticket for it too.

app/models/request.rb
```diff
@@ -15,7 +16,7 @@ class Request < ApplicationRecord
 
   scope :decision,  ->(decision)  { where(:decision => decision) }
   scope :state,     ->(state)     { where(:state => state) }
-  scope :requester, ->(requester) { where(:requester => requester) }
+  scope :owner,     ->(owner)     { where(:owner => owner) }
```
bzwei: reformat/align the spaces from line 17-19

db/migrate/20190612021005_rename_requester_to_owner.rb
```diff
@@ -0,0 +1,5 @@
+class RenameRequesterToOwner < ActiveRecord::Migration[5.2]
+  def change
+    rename_column :requests, :requester, :owner
```
bzwei: This migrate has problem to keep old data valid. The old requester column stored First and Last name while the new owner column stores username. I don't have better solution. Maybe it is OK to break backward compatibility.

public/doc/openapi-3-v1.0.0.json
```diff
@@ -1148,9 +1148,9 @@
                     "content"
                 ],
                 "properties": {
-                    "requester": {
+                    "owner": {
```
bzwei: remove owner from `RequestIn`

PR #142, opened 2019-06-27
fix smoke marker

zsadeh:


zsadeh:
@gmcculloug another fix for tests marker name. Need to merge both to master and stable

PR #144, opened 2019-06-27
Added RBAC access acls/roles to service

hsong-rh:
This is the second phase to enable RBAC support in Approval service:

1. create `RBAC::Access` to handle access control list (acls) based on predefined permissions and resource definitions;
2. set role (`admin, approver, owner`) according to acls;

### DIFF
--- a/lib/rbac/access.rb
+++ b/lib/rbac/access.rb
@@ -1,0 +1,115 @@
+module RBAC
+  class Access
+    attr_reader :acls
+    attr_reader :approver_acls
+    attr_reader :owner, :approver, :admin
+
+    def initialize(resource, verb)
+      @resource      = resource
+      @verb          = verb
+      @regexp        = Regexp.new(":(#{@resource}|\\*):(#{@verb}|\\*)")
+      @app_name      = ENV["APP_NAME"] || "approval"
+      @admin         = false
+      @approver      = false
+      @owner         = false
+      @acls          = []
+      @approver_acls = []
+    end
+
+    def process
+      RBAC::Service.call(RBACApiClient::AccessApi) do |api|
+        Rails.logger.info("Fetch access list for #{@app_name}")
+        @acls = RBAC::Service.paginate(api, :get_principal_access, {}, @app_name).select do |item|
+          Rails.logger.info("Found ACL: #{item}")
+          @regexp.match(item.permission)
+        end
+
+        @acls.any? do |item|
+          item.resource_definitions.any? do |rd|
+            @admin = true if rd.attribute_filter.key == 'id' &&
+                             rd.attribute_filter.operation == 'equal' &&
+                             rd.attribute_filter.value == '*'
+            @owner = true if rd.attribute_filter.key == 'owner' &&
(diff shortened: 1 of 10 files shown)


lib/rbac/permissions.rb
```diff
@@ -0,0 +1,8 @@
+module RBAC
+  module Permissions
```
bzwei: why there are only four permissions?

lib/rbac/access.rb
```diff
@@ -0,0 +1,123 @@
...
+        @acls.any? do |item|
+          item.resource_definitions.any? do |rd|
+            @admin = true if rd.attribute_filter.key == 'id' &&
+                             rd.attribute_filter.operation == 'equal' &&
+                             rd.attribute_filter.value == '*'
+          end
+        end
+
+        approver_regexp = Regexp.new(":(workflows|\\*):(approve|\\*)")
+        @approver_acls = RBAC::Service.paginate(api, :get_principal_access, {}, @app_name).select do |item|
```
bzwei: should not call `RBAC::Service` twice. Reuse the result from line 25

lib/rbac/permissions.rb
hsong-rh: Other permissions are predefined.

lib/rbac/acls.rb
```diff
@@ -0,0 +1,70 @@
...
+      end
+      acls
+    end
+
+    def add(acls, resource_id, permissions)
+      new_acls = []
+      permissions.each do |permission|
+        next if find_matching(acls, resource_id, permission)
+
+        new_acls << create_acl(permission, resource_id)
```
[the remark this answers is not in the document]
hsong-rh: Need to remove duplicates.

lib/rbac/access.rb
```diff
@@ -0,0 +1,161 @@
...
+
+    def process
+      RBAC::Service.call(RBACApiClient::AccessApi) do |api|
+        Rails.logger.info("Fetch access list for #{@app_name}")
+        full_acls = RBAC::Service.paginate(api, :get_principal_access, {}, @app_name)
+
+        approver_regexp = Regexp.new(":(workflows|\\*):(approve|\\*)")
+
+        full_acls.each do |item|
+          Rails.logger.info("Found ACL: #{item}")
```
bzwei: do you want to log every ACL? Perhaps debug is better.

lib/rbac/access.rb
```diff
@@ -0,0 +1,161 @@
...
+      @approver_acls = []
+      @owner_acls    = []
+    end
+
+    def process
+      RBAC::Service.call(RBACApiClient::AccessApi) do |api|
+        Rails.logger.info("Fetch access list for #{@app_name}")
+        full_acls = RBAC::Service.paginate(api, :get_principal_access, {}, @app_name)
+
+        approver_regexp = Regexp.new(":(workflows|\\*):(approve|\\*)")
```
bzwei: Create two regexes in one place. Either in constructor or here.

PR #145, opened 2019-06-28
Predefined RBAC roles/acls/policies/groups

hsong-rh:
This is the third phase to enable RBAC support in Approval service:

1. `rbac_approval_seed.yml` lists out all needed roles/acls/groups/policies for approval service; 
2. `rbac_approval_seed.yml` is also used to preconfigure RBAC environment (https://github.com/RedHatInsights/insights-rbac/pull/95);
3. provided a rake task to setup local RBAC environments;

### DIFF
--- a/data/rbac_approval_seed.yml
+++ b/data/rbac_approval_seed.yml
@@ -4,11 +4,6 @@ resource_definition1: &AdminResourceDefinition
     value: '*'
     key: id
     operation: equal
-resource_definition2: &OwnerResourceDefinition
-  attributeFilter: 
-    value: '{{username}}'
-    key: owner
-    operation: equal
 access1: &AdminReadAccessTemplates
   permission: approval:templates:read
   resourceDefinitions:
@@ -41,12 +36,6 @@ access8: &AdminCreateAccessActions
   permission: approval:actions:create
   resourceDefinitions:
           - *AdminResourceDefinition
-access51: &RequesterCreateAccessRequests
-  permission: approval:requests:create
-access52: &RequesterReadAccessRequests
-  permission: approval:requests:read
-  resourceDefinitions:
-          - *OwnerResourceDefinition
 access61: &ApproverCreateAccessActions
   permission: approval:actions:create
 access62: &ApproverReadAccessActions
@@ -69,15 +58,7 @@ role1: &ApprovalAdministratorRole
     - *AdminReadAccessStages
     - *AdminReadAccessActions
     - *AdminCreateAccessActions
-role2: &ApprovalRequesterRole
-  name: Approval Requester
-  system: true
-  version: 3
-  description: An Approval requester roles grants read and create permissions
-  access:
-    - *RequesterCreateAccessRequests
-    - *RequesterReadAccessRequests
-role3: &ApprovalApproverRole
+role2: &ApprovalApproverRole
   name: Approval Approver
   system: true
   version: 3
@@ -90,26 +71,15 @@ role3: &ApprovalApproverRole
 group1: &ApprovalAdministratorsGroup
   name: Approval Administrators
   description: Approval Administrators have complete access to all objects in the Approval Service.
-group2: &ApprovalRequestersGroup
-  name: Approval Requesters
-  description: Approval Requesters have limited access to certain objects in the Approval Service.
 policy1: &ApprovalAdministratorPolicy
   name: Approval Administrator
   group: *ApprovalAdministratorsGroup
   role: *ApprovalAdministratorRole
   description: Policy for Approval Administrators who have complete access to all objects in the Approval Service.
-policy2: &ApprovalRequesterPolicy
-  name: Approval Requester
-  group: *ApprovalRequestersGroup
-  role: *ApprovalRequesterRole
-  description: Policy for Approval Requesters who can limit access to certain objects in the Approval Service
 roles:
   - *ApprovalAdministratorRole
-  - *ApprovalRequesterRole
   - *ApprovalApproverRole
 groups:
   - *ApprovalAdministratorsGroup
-  - *ApprovalRequestersGroup
 policies:
   - *ApprovalAdministratorPolicy
-  - *ApprovalRequesterPolicy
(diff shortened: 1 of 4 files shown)


data/rbac_approval_seed.yml
```diff
@@ -0,0 +1,85 @@
...
+access62: &ApproverReadAccessActions
+  permission: approval:actions:read
+access63: &ApproverReadAccessStages
+  permission: approval:stages:read
+access64: &ApproverReadAccessRequests
+  permission: approval:requests:read
+role1: &ApprovalAdministratorRole
+  name: Approval Administrator
+  system: true
+  version: 3
```
mkanoor: @hsong-rh This should be version 1, its meant to control your data. It is away for the RBAC service to update the roles when a new version is released by your service/app.

lib/tasks/seed_rbac_data.rake
```diff
@@ -0,0 +1,11 @@
+require 'rake'
```
mkanoor: @hsong-rh Is this only for testing, since the role seeding would be done by the RBAC service. Even if we need a rake task it would only be for seeding the groups and roles, since the RBAC service currently doesn't provide a way to seed those.

lib/rbac/seed.rb
```diff
@@ -0,0 +1,129 @@
+require 'rbac-api-client'
+require 'pry'
```
mkanoor: @hsong-rh Is pry needed?

data/rbac_approval_seed.yml
```diff
@@ -0,0 +1,85 @@
...
+access6: &AdminReadAccessStages
+  permission: approval:stages:read
+  resourceDefinitions:
+          - *AdminResourceDefinition
+access7: &AdminReadAccessActions
+  permission: approval:actions:read
+  resourceDefinitions:
+          - *AdminResourceDefinition
+access8: &AdminCreateAccessActions
+  permission: approval:actions:create
```
mkanoor: @hsong-rh Should this be write instead of create, or do you want to keep create and update separate.

lib/tasks/seed_rbac_data.rake
hsong-rh: Yes, this is only for testing purpose.

lib/rbac/seed.rb
hsong-rh: Will remove it.

data/rbac_approval_seed.yml
hsong-rh: OK, will change to 1

data/rbac_approval_seed.yml
hsong-rh: Yes, we want to keep the difference between `create` and `update`. This is also the reason why we don't use `write`. So far `update` is only allowed on `workflow` for admin.

lib/rbac/seed.rb
```diff
@@ -0,0 +1,133 @@
...
+      end
+    end
+
+    def current_groups
+      RBAC::Service.call(RBACApiClient::GroupApi) do |api|
+        RBAC::Service.paginate(api, :list_groups,  {}).to_a
+      end
+    end
+
+    def create_roles
```
bzwei: why don't you use the `Roles` class you created under the same directory? Same for all other RBAC objects.

data/rbac_approval_seed.yml
```diff
@@ -0,0 +1,85 @@
...
+  permission: approval:actions:read
+access63: &ApproverReadAccessStages
+  permission: approval:stages:read
+access64: &ApproverReadAccessRequests
+  permission: approval:requests:read
+role1: &ApprovalAdministratorRole
+  name: Approval Administrator
+  system: true
+  version: 1
+  description: An Approval administrator roles grants create, read and update permissions
```
bzwei: An approval administrator role that grants create, read, and update permissions.

data/rbac_approval_seed.yml
```diff
@@ -0,0 +1,85 @@
...
+    - *AdminUpdateAccessWorkflows
+    - *AdminReadAccessRequests
+    - *AdminReadAccessStages
+    - *AdminReadAccessActions
+    - *AdminCreateAccessActions
+role2: &ApprovalApproverRole
+  name: Approval Approver
+  system: true
+  version: 1
+  description: An Approval approver roles grants read and create permissions
```
bzwei: An approver role that grants read and create permissions.

data/rbac_approval_seed.yml
```diff
@@ -0,0 +1,85 @@
...
+  access:
+    - *ApproverReadAccessRequests
+    - *ApproverReadAccessStages
+    - *ApproverReadAccessActions
+    - *ApproverCreateAccessActions
+group1: &ApprovalAdministratorsGroup
```
bzwei: &ApprovalAdministratorGroup

data/rbac_approval_seed.yml
```diff
@@ -0,0 +1,85 @@
...
+policy1: &ApprovalAdministratorPolicy
+  name: Approval Administrator
+  group: *ApprovalAdministratorsGroup
+  role: *ApprovalAdministratorRole
+  description: Policy for Approval Administrators who have complete access to all objects in the Approval Service.
+roles:
+  - *ApprovalAdministratorRole
+  - *ApprovalApproverRole
+groups:
+  - *ApprovalAdministratorsGroup
```
bzwei: *ApprovalAdministratorGroup

data/rbac_approval_seed.yml
```diff
@@ -0,0 +1,90 @@
...
+access62: &ApproverReadAccessActions
+  permission: approval:actions:read
+access63: &ApproverReadAccessStages
+  permission: approval:stages:read
+access64: &ApproverReadAccessRequests
+  permission: approval:requests:read
+role1: &ApprovalAdministratorRole
+  name: Approval Administrator
+  system: true
+  version: 1
```
bzwei: Don't you need version 2?

hsong-rh: yes.

PR #147, opened 2019-07-05
Add GraphQL support

bzwei:
https://projects.engineering.redhat.com/browse/SSP-543

depends on https://github.com/ManageIQ/manageiq-api-common/pull/86


bzwei:
@abellotti please review

abellotti:
looks about right, do you use the client generator ? if so can you run it against this updated openapi json to make sure the updated client's changes is as expected ?

Thanks,

Alberto

bzwei:
I tested with generated client. It works great.

PR #149, opened 2019-07-11
Add ability to query for default workflow

lindgrenj6:
https://projects.engineering.redhat.com/browse/SSP-535

For the catalog we need the ability to query for the default workflow due to the fact that when one isn't selected it just kind of sits in limbo. 

This will be used on the catalog tenant settings side as well as in the future when we have validation on the approval workflow the user selects.

### DIFF
--- a/app/controllers/api/v1x0/workflows_controller.rb
+++ b/app/controllers/api/v1x0/workflows_controller.rb
@@ -9,9 +9,12 @@ module Api
       end
 
       def show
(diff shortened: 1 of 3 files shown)


lindgrenj6:
@miq-bot add_reviewer bzwei 

cc @gmcculloug

lindgrenj6:
Closing as we don't need this functionality currently, we may in the future but as of right now it isn't needed.

PR #151, opened 2019-07-17
Added services to create RBAC roles based on workflow's groups

hsong-rh:
This is the fourth PR to enable RBAC support in Approval service:

Create or update RBAC roles when a workflow with groups information is created or updated.

### DIFF
--- a/app/services/access_process_service.rb
+++ b/app/services/access_process_service.rb
@@ -2,7 +2,7 @@ class AccessProcessService
   include RBAC::Permissions
 
   def initialize(opts = {})
-    @app_name = opts[:app_name] || 'approval'
+    @app_name = ENV["APP_NAME"] || 'approval'
     @prefix = opts[:role_prefix] || "#{@app_name}-group-"
     @acls = RBAC::ACLS.new
     @roles = RBAC::Roles.new(@prefix)
--- a/app/services/access_process_service.rb
+++ b/app/services/access_process_service.rb
@@ -6,7 +6,24 @@ class AccessProcessService
     @prefix = opts[:role_prefix] || "#{@app_name}-group-"
     @acls = RBAC::ACLS.new
     @roles = RBAC::Roles.new(@prefix)
+    @policies = RBAC::Policies.new(@prefix)
   end
+
+  def add_resource_to_groups(resource_id, group_refs, permissions = [WORKFLOW_APPROVE_PERMISSION])
+    group_refs.each do |uuid|
+      name = "#{@prefix}#{uuid}"
+      update_or_create_role(name, resource_id, permissions)
+    end
+  end
+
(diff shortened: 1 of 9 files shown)


app/services/access_process_service.rb
```diff
@@ -0,0 +1,95 @@
...
+  def create_role(name, resource_id, permissions)
+    acls = @acls.create(resource_id, permissions)
+    role = @roles.add(name, acls)
+
+    add_policy(name, role.uuid)
+
+    role
+  end
+
+  def add_role_acls(role, resource_id, permissions = [WORKFLOW_APPROVE_PERMISSION])
```
bzwei: permissions does not need default value

app/services/access_process_service.rb
```diff
@@ -0,0 +1,95 @@
...
+
+  private
+
+  def create_role(name, resource_id, permissions)
+    acls = @acls.create(resource_id, permissions)
+    role = @roles.add(name, acls)
+
+    add_policy(name, role.uuid)
+
+    role
```
bzwei: why do you need to return `role`? wrap them into a `tap` block?

app/services/access_process_service.rb
```diff
@@ -0,0 +1,95 @@
...
+      policy_in = RBACApiClient::PolicyIn.new
+      policy_in.name = name
+      policy_in.description = 'Approval Policy'
+      policy_in.group = name.split('group-')[1]
+      policy_in.roles = [role_uuid]
+      api_instance.create_policies(policy_in)
+    end
+  end
+
+  def delete_policy(role)
```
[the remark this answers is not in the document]
hsong-rh: The role we created is always bound to the group. So the relationship between policy and role is always one-to-one.

app/services/access_process_service.rb
```diff
@@ -0,0 +1,75 @@
...
+    role = find_role(name)
+
+    role ? add_role_acls(role, resource_id, permissions) : create_role(name, resource_id, permissions)
+  end
+
+  def find_role(name)
+    @roles.find(name)
+  end
+
+  def add_resource_to_groups(resource_id, group_refs)
```
bzwei: It seems `add_resource_to_groups` and `remove_resource_from_groups` are the only two methods used outside. Add `permissions = [WORKFLOW_APPROVE_PERMISSION]` in the arg list and turn other public methods into private.

app/services/access_process_service.rb
```diff
@@ -0,0 +1,75 @@
+class AccessProcessService
+  include RBAC::Permissions
+
+  def initialize(opts = {})
+    @app_name = opts[:app_name] || 'approval'
```
mkanoor: @hsong-rh Could this be ENV['APP_NAME']
https://github.com/ManageIQ/catalog-api/blob/ed1d457e9353aae515684d2445126e749d621088/lib/rbac/access.rb#L9

PR #152, opened 2019-07-18
Added rbac to resource index action

hsong-rh:
This is the fifth PR to enable RBAC support in Approval service: added RBAC checking on resource index action.

### DIFF
--- a/app/controllers/api/v1x0/actions_controller.rb
+++ b/app/controllers/api/v1x0/actions_controller.rb
@@ -36,11 +36,10 @@ module Api
       end
 
       def rbac_scope(relation)
-        access_obj = RBAC::Access.new('actions', 'read').process
+        access_obj = rbac_read_access(relation)
         return relation if access_obj.admin?
 
-        raise Exceptions::NotAuthorizedError, "Not Authorized to list actions" unless access_obj.approver? || access_obj.accessible?
-
+        # Only approver can reach here
         action_ids = access_obj.approver_id_list
         Rails.logger.info("approver scope for actions: #{action_ids}")
 
--- a/app/controllers/api/v1x0/actions_controller.rb
+++ b/app/controllers/api/v1x0/actions_controller.rb
@@ -4,7 +4,8 @@ module Api
       include Mixins::IndexMixin
       def index
         stage = Stage.find(params.require(:stage_id))
-        collection(stage.actions)
+
+        RBAC::Access.enabled? ? collection(rbac_scope(stage.actions)) : collection(stage.actions)
       end
 
       def show
@@ -15,14 +16,14 @@ module Api
 
       def create
         stage_id = if params[:request_id]
-          req = Request.find(params[:request_id])
-          current_stage = req.current_stage
-          raise Exceptions::ApprovalError, "Request has finished its lifecycle. No more action can be added to its current stage." unless current_stage
+                     req = Request.find(params[:request_id])
+                     current_stage = req.current_stage
+                     raise Exceptions::ApprovalError, "Request has finished its lifecycle. No more action can be added to its current stage." unless current_stage
 
-          current_stage.id
-        else
-          params.require(:stage_id)
-        end
+                     current_stage.id
+                   else
+                     params.require(:stage_id)
+                   end
 
         action = ActionCreateService.new(stage_id).create(action_params)
         json_response(action, :created)
@@ -32,6 +33,18 @@ module Api
 
       def action_params
         params.permit(:operation, :processed_by, :comments)
+      end
+
+      def rbac_scope(relation)
+        access_obj = RBAC::Access.new('actions', 'read').process
+        return relation if access_obj.admin?
+
+        raise Exceptions::NotAuthorizedError, "Not Authorized to list actions" unless access_obj.approver? || access_obj.accessible?
+
+        action_ids = access_obj.approver_id_list
+        Rails.logger.info("approver scope for actions: #{action_ids}")
+
+        relation.where(:id => action_ids)
       end
     end
   end
--- a/app/controllers/api/v1x0/mixins/index_mixin.rb
+++ b/app/controllers/api/v1x0/mixins/index_mixin.rb
@@ -26,6 +26,13 @@ module Api
           ManageIQ::API::Common::Filter.new(base_query, params[:filter], api_doc_definition).apply
         end
 
+        def rbac_read_access(relation)
+          access_obj = RBAC::Access.new(relation.model.table_name, 'read').process
+          raise Exceptions::NotAuthorizedError, "Not Authorized to list #{relation.model.table_name}" unless access_obj.accessible?
+
+          access_obj
+        end
+
         private
 
         def api_doc_definition
--- a/app/controllers/api/v1x0/requests_controller.rb
+++ b/app/controllers/api/v1x0/requests_controller.rb
@@ -32,37 +32,34 @@ module Api
       end
 
       def rbac_scope(relation)
-        access_obj = RBAC::Access.new('requests', 'read').process
-        raise Exceptions::NotAuthorizedError, "Not Authorized to list requests" unless access_obj.accessible?
+        access_obj = rbac_read_access(relation)
+
+        return relation if access_obj.admin?
 
         # return error for using wrong path
         raise Exceptions::NotAuthorizedError, "Current role cannot access #{request.path}" unless right_path?(access_obj)
 
-        return relation if access_obj.admin?
+        ids = if approver_endpoint?
+                access_obj.approver_id_list
+              else
+                access_obj.owner_id_list
+              end
 
-        approver_id_list = access_obj.approver_id_list
-        approver_relation = relation.where(:id => approver_id_list)
-        Rails.logger.info("approver scope for requests: #{approver_id_list}")
+        Rails.logger.info("Accessible request ids: #{ids}")
 
-        owner_id_list = access_obj.owner_id_list
-        owner_relation = relation.where(:id => owner_id_list)
-        Rails.logger.info("Owner scope for requests: #{owner_id_list}")
-
-        # double roles for requests
-        if approver_relation.any? && owner_relation.any?
-          return request.path.end_with?("/approver/requests") ? approver_relation : owner_relation
-        end
-
-        # For other resources
-        return approver_relation if approver_relation.any?
-
-        owner_relation
+        relation.where(:id => ids)
       end
 
       def right_path?(access_obj)
-        (access_obj.approver? && request.path.end_with?("/approver/requests")) ||
-          (access_obj.owner? && request.path.end_with?("/requester/requests")) ||
-          (access_obj.admin? && !request.path.end_with?("/approver/requests", "/requester/requests"))
+        (access_obj.approver? && approver_endpoint?) || (access_obj.owner? && requester_endpoint?)
+      end
+
+      def requester_endpoint?
+        request.path.end_with?("/requester/requests")
+      end
+
+      def approver_endpoint?
+        request.path.end_with?("/approver/requests")
       end
     end
   end
--- a/app/controllers/api/v1x0/requests_controller.rb
+++ b/app/controllers/api/v1x0/requests_controller.rb
@@ -22,13 +22,47 @@ module Api
                  Request.includes(:stages)
                end
 
-        collection(reqs)
+        RBAC::Access.enabled? ? collection(rbac_scope(reqs)) : collection(reqs)
       end
 
       private
 
       def request_params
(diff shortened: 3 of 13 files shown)


app/controllers/api/v1x0/stages_controller.rb
```diff
@@ -10,7 +10,22 @@ def show
...
-        collection(req.stages)
+
+        RBAC::Access.enabled? ? collection(rbac_scope(req.stages)) : collection(req.stages)
+      end
+
+      private
+
+      def rbac_scope(relation)
+        access_obj = RBAC::Access.new('stages', 'read').process
+        raise Exceptions::NotAuthorizedError, "Not Authorized to list stages" unless access_obj.accessible? || access_obj.approver?
```
bzwei: `approver?` should  be `owner?`, right?

app/controllers/api/v1x0/actions_controller.rb
```diff
@@ -33,6 +34,18 @@ def create
       def action_params
         params.permit(:operation, :processed_by, :comments)
       end
+
+      def rbac_scope(relation)
+        access_obj = RBAC::Access.new('actions', 'read').process
+        return relation if access_obj.admin?
+
+        raise Exceptions::NotAuthorizedError, "Not Authorized to list actions" unless access_obj.approver? || access_obj.accessible?
```
bzwei: Mustn't approver have accessible right? So check for `assessable?` should be sufficient.

app/controllers/api/v1x0/actions_controller.rb
```diff
@@ -33,6 +34,18 @@ def create
...
+
+        action_ids = access_obj.approver_id_list
```
bzwei: Please add a comment here saying only approver is allowed up to here

app/controllers/api/v1x0/requests_controller.rb
```diff
@@ -22,14 +22,48 @@ def index
...
+          return request.path.end_with?("/approver/requests") ? approver_relation : owner_relation
+        end
+
+        # For other resources
+        return approver_relation if approver_relation.any?
+
+        owner_relation
+      end
+
+      def right_path?(access_obj)
```
bzwei: refactor all `end_with?` calls into methods such as `approver_endpoint?`, `owner_endpoint?` etc. we may not need `admin_endpoint?`

app/controllers/api/v1x0/requests_controller.rb
```diff
@@ -22,14 +22,48 @@ def index
...
       def request_params
         params.permit(:name, :description, :requester_name, :content => {})
       end
+
+      def rbac_scope(relation)
+        access_obj = RBAC::Access.new('requests', 'read').process
+        raise Exceptions::NotAuthorizedError, "Not Authorized to list requests" unless access_obj.accessible?
+
+        # return error for using wrong path
+        raise Exceptions::NotAuthorizedError, "Current role cannot access #{request.path}" unless right_path?(access_obj)
```
bzwei: move line 41 up. Then it is sufficient here to only check for approver path and owner path.

app/controllers/api/v1x0/requests_controller.rb
```diff
@@ -22,14 +22,48 @@ def index
...
+
+        return relation if access_obj.admin?
+
+        approver_id_list = access_obj.approver_id_list
```
bzwei: Can't we rewrite the following logic this way?
```
relation =
  if approver_endpoint?
    relation.where(:id => access_obj.approval_id_list)
  else
    relation.where(:id => access_obj.owner_id_list)
  end
Rails.logger.info(...)
relation
```
I don't see the case for double roles

app/controllers/api/v1x0/stages_controller.rb
```diff
@@ -10,7 +10,22 @@ def show
...
+        RBAC::Access.enabled? ? collection(rbac_scope(req.stages)) : collection(req.stages)
+      end
+
+      private
+
+      def rbac_scope(relation)
+        access_obj = RBAC::Access.new('stages', 'read').process
+        return relation if access_obj.admin?
+
+        raise Exceptions::NotAuthorizedError, "Not Authorized to list stages" unless access_obj.owner? || access_obj.accessible?
```
bzwei: Same as actions controller. Check only for `accessible?` and comment below only `owner` is allowed.

app/controllers/api/v1x0/templates_controller.rb
```diff
@@ -10,7 +10,17 @@ def show
...
-        collection(templates)
+
+        RBAC::Access.enabled? ? collection(rbac_scope(templates)) : collection(templates)
+      end
+
+      private
+
+      def rbac_scope(relation)
+        access_obj = RBAC::Access.new('templates', 'read').process
+        raise Exceptions::NotAuthorizedError, "Not Authorized to list templates" unless access_obj.accessible? || access_obj.admin?
```
bzwei: Check only `accessible?`

app/controllers/api/v1x0/workflows_controller.rb
```diff
@@ -48,6 +49,13 @@ def update
       def workflow_params
         params.permit(:name, :description, :group_refs => [])
       end
+
+      def rbac_scope(relation)
+        access_obj = RBAC::Access.new('workflows', 'read').process
+        raise Exceptions::NotAuthorizedError, "Not Authorized to list workflows" unless access_obj.accessible? || access_obj.admin?
```
bzwei: check only `accessible?`. combine line 54 and 55 into one line. Maybe rubocop will no longer complain for code similarity.

lib/rbac/access.rb
```diff
@@ -26,16 +26,17 @@ def process
         full_acls.each do |item|
           Rails.logger.debug("Found ACL: #{item}")
           @acls << item if regexp.match(item.permission)
+          @approver_acls << item if approver_regexp.match(item.permission)
+        end
 
-          unless @admin
+        unless @admin
```
bzwei: I don't understand the changes in this file. The old code looks fine to me.

spec/controllers/v1.0/requests_controller_spec.rb
```diff
@@ -40,49 +48,111 @@
...
+      allow(access_obj).to receive(:process).and_return(access_obj)
+      allow(access_obj).to receive(:approver_id_list).and_return([])
+      allow(access_obj).to receive(:owner_id_list).and_return([])
 
-    it 'returns requests' do
-      expect(json['links']).not_to be_empty
-      expect(json['links']['first']).to match(/limit=5&offset=0/)
-      expect(json['links']['last']).to match(/limit=5&offset=5/)
-      expect(json['data'].size).to eq(5)
+      get "#{api_version}/requests", :params => { :limit => 5, :offset => 0 }, :headers => request_header
```
bzwei: in general all `get` commands should be placed inside `it` block rather in `before` block. You may want to create a separate PR to do the move all for occurrences.

lib/rbac/access.rb
hsong-rh: Old one loops on all acls, i.e `full_acls`. The right way should only loop on `acls`.

app/controllers/api/v1x0/actions_controller.rb
```diff
@@ -33,6 +34,19 @@ def create
       def action_params
         params.permit(:operation, :processed_by, :comments)
       end
+
+      def rbac_scope(relation)
+        access_obj = RBAC::Access.new('actions', 'read').process
```
bzwei: with this round of refactoring, now we can extract the first three lines into one method that can be shared by all controllers. This method returns `access_obj` if we need to further narrow the scope. Return `nil` if it is admin.

app/controllers/api/v1x0/mixins/index_mixin.rb
```diff
@@ -26,6 +26,13 @@ def filtered(base_query)
           ManageIQ::API::Common::Filter.new(base_query, params[:filter], api_doc_definition).apply
         end
 
+        def rbac_access(relation)
```
bzwei: consider rename to `rbac_read_access`

lib/rbac/access.rb
```diff
@@ -25,13 +25,14 @@ def process
...
-
-          unless @admin
-            item.resource_definitions.any? do |rd|
-              @admin = true if rd.attribute_filter.key == 'id' &&
-                               rd.attribute_filter.operation == 'equal' &&
-                               rd.attribute_filter.value == '*'
+          if regexp.match(item.permission)
+            @acls << item
+            unless @admin
+              item.resource_definitions.any? do |rd|
```
mkanoor: @hsong-rh can this admin check be changed to be role specific. We are working with the RBAC team to come api with an API that would allow roles to be fetched without having to check for individual accesses.
https://github.com/RedHatInsights/insights-rbac/issues/115
We can merge this for now but you might want to open a GitHub issue to convert this to be role based.
Better yet if we can extract this into a function we can just swap out the logic to role based instead of access permission based.

spec/controllers/v1.0/requests_controller_spec.rb
```diff
@@ -40,49 +48,111 @@
...
+
+    context 'when approver' do
+      let(:access_obj) { instance_double(RBAC::Access, :accessible? => true, :admin? => false, :approver? => true, :owner? => false) }
+      before do
+        allow(RBAC::Access).to receive(:new).with('requests', 'read').and_return(access_obj)
+        allow(access_obj).to receive(:process).and_return(access_obj)
+      end
+
+      it 'returns status code 403' do
+        get "#{api_version}/workflows/#{workflow_id}/requests", :headers => request_header
```
mkanoor: @hsong-rh can the request_header be moved into a spec support file and shared across all tests. If we control it from a single file we can add new keys
https://github.com/ManageIQ/catalog-api/blob/cee85e285bec837b21cc0fe6723097bdfbd2539c/spec/support/service_spec_helper.rb#L7

spec/controllers/v1.0/requests_controller_spec.rb
mkanoor: @hsong-rh do we need a limit and offset here shouldn't the defaults work properly?

spec/controllers/v1.0/requests_controller_spec.rb
```diff
@@ -166,17 +255,64 @@
     let(:item) { { 'disk' => '100GB' } }
     let(:valid_attributes) { { :requester_name => '1234', :name => 'Visit Narnia', :content => item, :description => 'desc' } }
 
-    context 'when request attributes are valid' do
+    context 'admin role when request attributes are valid' do
+      let(:access_obj) { instance_double(RBAC::Access, :accessible? => true, :admin? => true, :approver? => false, :owner? => false) }
+      before do
+        ENV['AUTO_APPROVAL'] = 'y'
```
mkanoor: @hsong-rh Instead of setting the environment variable directly can we use this approach
https://github.com/ManageIQ/catalog-api/blob/cee85e285bec837b21cc0fe6723097bdfbd2539c/spec/lib/dev_credentials_spec.rb#L4
Using the gem 'climate_control'

spec/controllers/v1.0/requests_controller_spec.rb
```diff
@@ -166,17 +255,64 @@
...
+
+      it 'returns status code 500' do
+        expect(response).to have_http_status(500)
+      end
+    end
+
+    context 'approver role' do
+      let(:access_obj) { instance_double(RBAC::Access, :accessible? => true, :admin? => false, :approver? => true, :owner? => false) }
+      before do
+        ENV['AUTO_APPROVAL'] = 'y'
```
mkanoor: @hsong-rh can we use the with_modified_env instead of setting the ENV variables directly

spec/controllers/v1.0/templates_controller_spec.rb
```diff
@@ -11,18 +11,51 @@
...
+        expect(json['links']['last']).to match(/limit=5&offset=5/)
+        expect(json['data'].size).to eq(5)
+        expect(response).to have_http_status(200)
+      end
+    end
+
+    context 'when approver role' do
+      let(:access_obj) { instance_double(RBAC::Access, :accessible? => false, :admin? => false, :approver? => true, :owner? => false) }
+
+      before { get "#{api_version}/templates", :params => { :limit => 5, :offset => 0 }, :headers => request_header }
```
mkanoor: @hsong-rh do we need the limit and offset shouldn't that just use the defaults.

lib/rbac/access.rb
hsong-rh: @mkanoor Issue https://github.com/ManageIQ/approval-api/issues/159 is created.

PR #154, opened 2019-07-29
Added a rake task to sync RBAC roles for existing workflows

hsong-rh:
`rake approval:sync_rbac_roles` task is added to sync RBAC roles for workflows.

### DIFF
--- a/app/controllers/api/v1x0/workflows_controller.rb
+++ b/app/controllers/api/v1x0/workflows_controller.rb
@@ -38,7 +38,7 @@ module Api
 
       def update
         workflow = Workflow.find(params.require(:id))
-        workflow.update(workflow_params)
+        WorkflowUpdateService.new(workflow.id).update(workflow_params)
 
         json_response(workflow)
       end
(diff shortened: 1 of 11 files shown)


mkanoor:
@hsong-rh What is the use case for this is this only for development and QE as part of debugging cleanup. In real production environment we won't be able to run any rake tasks.

hsong-rh:
@mkanoor After RBAC is enabled, approver in a certain group cannot process any requests assigned  to him, because the group he belongs to does not have role. This rake task will create roles for all such groups.

mkanoor:
@hsong-rh After these PR's are merged wouldn't RBAC be enabled as default so this rake task is a temporary thing for QE and DEV correct?

bzwei:
@mkanoor It is true the tool is a one time and temporary thing. It does not need to be a rake task. We just need a way to share the tool and save the source code. Any suggestion?

mkanoor:
@bzwei @hsong-rh I think any rake task is a hack and not practical in a production containerized environment. We have a similar set to debugging rake tasks that we carry in catalog service. So for now it's ok. Maybe at some point we should start deprecating it.

bzwei:
we can do the same thing. We can remove the task after everybody enrolls in RBAC roles.

PR #155, opened 2019-07-30
Keep roles in sync after role creation/deletion

hsong-rh:
Updated local variable `@roles` when role is created or deleted.


hsong-rh:
@bzwei Please review and merge

PR #157, opened 2019-08-08
Removed created request if errors happened afterwards

hsong-rh:
Currently a request will be created and left over in database even following RBAC or KIE service failed. This PR will handle this scenario.


hsong-rh:
@bzwei Please review.

bzwei:
@hsong-rh I felt it would be cleaner to use rails transaction for this purpose. The rollback will be automatically upon exception.

PR #158, opened 2019-08-08
Validate groups with roles before creating request

hsong-rh:
This PR added a logic to make sure that the groups used to create a request should have approval roles, otherwise will raise an exception.


app/models/group.rb
```diff
@@ -22,6 +22,15 @@ def self.all(username = nil)
     groups
   end
 
+  def with_approval_role?
```
mkanoor: @hsong-rh Could this be redone based on this issue https://github.com/RedHatInsights/insights-rbac/issues/115
To use the role names directly

hsong-rh: I don't think so, RedHatInsights/insights-rbac#115 will return roles based on users. Here I need check specific approval roles for groups.

app/services/jbpm_process_service.rb
```diff
@@ -39,6 +39,12 @@ def process_options
     options
   end
 
+  def validate(group)
+    raise Exceptions::RBACError, "Group #{group.uuid} doesn't have needed roles" unless group.with_approval_role?
```
bzwei: Can we print group name (or both) instead? The text can be straight `Group does not have approval role".

bzwei:
@hsong-rh can this PR be refactored with the new rbac api features?

bzwei:
not valid because of #243

PR #165, opened 2019-08-29
Refactor to use new RBAC APIs

hsong-rh:


lib/rbac/access.rb
```diff
@@ -1,123 +1,83 @@
...
-
-          @approver_acls << item if approver_regexp.match(item.permission)
-        end
-
-        @approver = true if approver_acls.any?
-
-        @owner_acls = requester_acls.select do |item|
-          regexp.match(item.permission)
-        end
+    ADMIN_ROLE = 'Approval Administrator'.freeze
```
mkanoor: @hsong-rh Since this is shared code should we put anything specific to Approval in here. At some point this is going to be in the shared common gem. The Admin would be different for Catalog/Approval/Topology and other service.

app/controllers/api/v1x0/requests_controller.rb
```diff
@@ -36,26 +36,25 @@ def request_params
...
-                access_obj.owner_id_list
+                { :id => RBAC::Access.approver_id_list(relation.model.table_name) }
+              elsif requester_endpoint?
+                { :id => RBAC::Access.owner_id_list(relation.model.table_name) }
               end
 
         Rails.logger.info("Accessible request ids: #{ids}")
 
-        relation.where(:id => ids)
+        relation.where(ids)
```
bzwei: I prefer the previous style `:id => ids` and keep `ids` as an array

lib/rbac/access.rb
```diff
@@ -1,123 +1,83 @@
...
+    APPROVER_ROLE = 'Approval Approver'.freeze
+    APP_NAME = 'approval'.freeze
 
-        @owner = true if @owner_acls.any?
-      end
-
-      Rails.logger.info("Role: admin[#{@admin}], approver[#{@approver}], owner[#{@owner}]")
-
-      self
+    def self.enabled?
```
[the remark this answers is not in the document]
hsong-rh: All class methods defined here are querying access lists, they should be thread safe.

lib/rbac/access.rb
hsong-rh: Will change when the real sharing starts.

public/doc/openapi-3-v1.0.0.json
```diff
@@ -274,6 +274,177 @@
                 }
             }
         },
+        "/admin/requests": {
```
bzwei: create a separate PR for this file

lib/rbac/approval_access.rb
```diff
@@ -0,0 +1,133 @@
+module RBAC
+  class ApprovalAccess < Access
+    include RBAC::Permissions
+
+    ADMIN_ROLE = 'Approval Administrator'.freeze
```
mkanoor: @hsong-rh To make this generic we should keep these constants in your controller/mixin and call generic functions in here. This file shouldn't contain anything approval specific but rather generic function which can be fed in values.

lib/rbac/approval_access.rb
```diff
@@ -0,0 +1,133 @@
...
+    end
+
+    # Request ids owned by requester
+    def owner_request_ids
+      Request.by_owner.pluck(:id).sort
+    end
+
+    # Request ids approver can process
+    def approver_request_ids
+      Request.where(:workflow_id => workflow_ids).pluck(:id).sort
```
mkanoor: @hsong-rh All ActiveRecord access should be in your mixin and not in the lib/....

lib/rbac/policies.rb
```diff
@@ -18,7 +18,7 @@ def add_policy(name, role_uuid)
     # delete all policies that contains the role.
     def delete_policy(role)
       RBAC::Service.call(RBACApiClient::PolicyApi) do |api_instance|
-        RBAC::Service.paginate(api_instance, :list_policies, :name => @prefix).each do |policy|
+        RBAC::Service.paginate(api_instance, :list_policies, :name => @prefix, :limit => 500).each do |policy|
```
mkanoor: @hsong-rh Should this be using the DEFAULT_LIMIT that is being defined in the lib/rbac/access.rb?

app/controllers/api/v1x0/mixins/rbac_mixin.rb
```diff
@@ -24,18 +33,108 @@ def read_access_check
         def resource_check(verb, id = params[:id], klass = controller_name.classify.constantize)
           return unless RBAC::Access.enabled?
 
-          access_obj = permission_check(verb, klass)
-
-          raise Exceptions::NotAuthorizedError, "#{verb.titleize} access not authorized for #{klass}" if !access_obj.admin? && access_obj.not_owned?(id.to_i) && access_obj.not_approvable?(id.to_i)
+          raise Exceptions::NotAuthorizedError, "#{verb.titleize} access not authorized for #{klass} with id: #{id}" unless resource_instance_accessible?(klass.table_name, id)
```
bzwei: why `resource_instance_accessible?` does not take `verb` as an argument?

app/controllers/api/v1x0/mixins/rbac_mixin.rb
```diff
@@ -24,18 +33,108 @@ def read_access_check
         def resource_check(verb, id = params[:id], klass = controller_name.classify.constantize)
```
bzwei: does this method work for `Workflow` and `Template`?

app/controllers/api/v1x0/mixins/rbac_mixin.rb
```diff
@@ -24,18 +33,108 @@ def read_access_check
...
           return unless RBAC::Access.enabled?
 
-          access_obj = permission_check(verb, klass)
-
-          raise Exceptions::NotAuthorizedError, "#{verb.titleize} access not authorized for #{klass}" if !access_obj.admin? && access_obj.not_owned?(id.to_i) && access_obj.not_approvable?(id.to_i)
+          raise Exceptions::NotAuthorizedError, "#{verb.titleize} access not authorized for #{klass} with id: #{id}" unless resource_instance_accessible?(klass.table_name, id)
         end
 
         def permission_check(verb, klass = controller_name.classify.constantize)
```
bzwei: does this method work for `Workflow` and `Template`?

app/controllers/api/v1x0/workflows_controller.rb
```diff
@@ -4,7 +4,7 @@ class WorkflowsController < ApplicationController
       include Mixins::IndexMixin
       include Mixins::RBACMixin
 
-      before_action :read_access_check, :only => %i[show]
+      before_action :index_access_check, :only => %i[show]
```
bzwei: should it be `:only => %i[show index]`?

app/controllers/api/v1x0/workflows_controller.rb
```diff
@@ -52,14 +52,19 @@ def update
...
 
-      def workflow_params
-        params.permit(:name, :description, :group_refs => [])
-      end
-
       def rbac_scope(relation)
-        rbac_read_access(relation)
+        return relation if admin?
+
+        # Only approver can reach here
```
bzwei: if only admin can see workflows, remove the rest

lib/rbac/roles.rb
```diff
@@ -16,6 +16,17 @@ def with_each_role
       end
     end
 
+    def self.assigned_role?(role_name)
```
bzwei: Should we have a method to get all role names in an array? In the code we have `admin?` and `approver?` that will required to make RBAC service call twice.

PR #167, opened 2019-09-11
Change external_tenant to string

syncrou:
This PR modifies the external_tenant attribute from an integer to a string.


gmcculloug:
@syncrou Thanks for the PR.  Do these test failures make sense?

syncrou:
@gmcculloug failures are resolved. @bzwei @hsong-rh would you mind taking a look?

PR #168, opened 2019-09-13
Removed logic to list requests by approver via query parameter

hsong-rh:
Remove approver filtering parameter in endpoint `/requests` in favor of the new endpoint `/requester/requests` that was added recently.

https://projects.engineering.redhat.com/browse/SSP-475


mkanoor:
@bzwei @hsong-rh 
Can an approver log into the system as a regular user to order items? When he does that he doesn't need to see things that he needs to approve.

hsong-rh:
@mkanoor Yes, approver can login as regular user to create request. But in this case, he has dual roles, means he is able to see those requests waiting for him to process. The request list is provided by different endpoints: `approver/requests` for approver role, and `requester/requests` for regular user.

PR #172, opened 2019-09-16
Corrected return code for Action API with invalid content

hsong-rh:
Changed return code to 400 (bad request) error code when posting an action with invalid content:
https://projects.engineering.redhat.com/browse/SSP-502

### DIFF
--- a/app/controllers/concerns/exception_handler.rb
+++ b/app/controllers/concerns/exception_handler.rb
@@ -19,7 +19,7 @@ module ExceptionHandler
       json_response({ :message => e.message }, :forbidden)
     end
 
-    rescue_from Exceptions::UserError do |e|
+    rescue_from Exceptions::UserError, Exceptions::ApprovalError do |e|
       json_response({ :message => e.message }, :bad_request)
     end
 
--- a/app/controllers/api/v1x0/stageaction_controller.rb
+++ b/app/controllers/api/v1x0/stageaction_controller.rb
@@ -6,9 +6,14 @@ module Api
 
       protect_from_forgery :with => :exception, :prepend => true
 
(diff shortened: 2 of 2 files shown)


hsong-rh:
@bzwei Please review

app/controllers/concerns/exception_handler.rb
```diff
@@ -19,7 +19,18 @@ module ExceptionHandler
       json_response({ :message => e.message }, :forbidden)
     end
 
-    rescue_from Exceptions::UserError do |e|
+    rescue_from RBACApiClient::ApiError do |e|
```
bzwei: I don't think we need to analyze any exception from an external system.

app/controllers/concerns/exception_handler.rb
```diff
@@ -19,7 +19,18 @@ module ExceptionHandler
...
+      when 403
+        json_response({ :message => e.message }, :forbidden)
+      when 404
+        json_response({ :message => e.message }, :not_found)
+      else
+        json_response({ :message => e.message }, :bad_request)
+      end
+    end
+
+    rescue_from Exceptions::UserError, Exceptions::ApprovalError do |e|
```
bzwei: Should we rescue all `ApprovalError` or only selected ones?

hsong-rh: All `ApprovalError` raised so far are suitable to category into bad requests.

app/controllers/concerns/exception_handler.rb
hsong-rh: Ok, then we will keep to return error code 500 whenever RBAC service raises exceptions.

PR #173, opened 2019-09-17
Return workflow when RBAC flag is turned off

hsong-rh:


app/controllers/api/v1x0/workflows_controller.rb
```diff
@@ -15,12 +15,11 @@ def create
 
       # TODO: remove 'approval:workflows:read' from approver acls list in RBAC Insight
       def show
-        return unless RBAC::Access.enabled?
+        workflow = Workflow.find(params.require(:id))
```
bzwei: how about?
```
def show
  if RBAC::Access.enabled?
    raise Exceptions::NotAuthorizedError, "Not Authorized for workflows" unless admin?
  end
  json_response(Workflow.find(params.require(:id))
end
```

Consider add a spec test.

PR #174, opened 2019-09-19
Added scope option when listing roles

hsong-rh:
Added scope option when initializing RBAC roles for two use cases:
1. use `:scope => 'principal'` when checking the roles for user;
2. use `:scope => 'account'` when creating approval roles for groups if needs;

https://projects.engineering.redhat.com/browse/SSP-776


bzwei:
Need some description and a Jira ticket

hsong-rh:
@bzwei SSP-776 is created for this issue.

PR #177, opened 2019-09-26
Updated openapi spec and check roles when post actions

hsong-rh:
1. Corrected spec for endpoints:`stages/{id}` and `/requests/{request_id}/actions`;
2. Added role checking before posting actions:
      `admin`:     can create all kinds of actions
      `approver`:  can not create 'cancel' action
      `requester`: can only create 'cancel' action

### DIFF
--- a/app/controllers/api/v1x0/actions_controller.rb
+++ b/app/controllers/api/v1x0/actions_controller.rb
@@ -5,7 +5,7 @@ module Api
       include Mixins::RBACMixin
 
       before_action :read_access_check, :only => %i[show]
-      before_action :create_access_check, :only => %i[create]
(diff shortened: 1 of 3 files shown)


app/controllers/api/v1x0/actions_controller.rb
```diff
@@ -36,6 +36,20 @@ def create
 
       private
 
+      def validate_create_action
+        raise Exceptions::NotAuthorizedError, "Not authorized to create [#{params[:operation]}] action " unless valid_operation
```
bzwei: do we really need to split the logic into two methods?

hsong-rh: Changed into one method

PR #178, opened 2019-09-27
Remove persona specific endpoint and take persona in header

bzwei:
https://projects.engineering.redhat.com/browse/SSP-794

Since each login may have multiple personas, we introduced three persona specific endpoints for requests, namely
/admin/requests
/approver/requests
/requester/requests

However this causes problem for GraphQL query because the persona is not a built in attributes of requests. Now we decide to remove these endpoints and take the persona parameter from header. The mapping from old endpoint to new is as follows
/admin/requests -> /requests with header x-rh-persona=approval/admin
/approver/requests -> /requests with header x-rh-persona=approval/approver
/requester/requests -> /requests with header x-rh-persona=approval/requester

A request without header x-rh-persona is the same as approval/requester
A request with unknown x-rh-persona header gets return code 403

The persona header works with RBAC together. A mismatching of persona and RBAC role results in a return code 403.

### DIFF
--- a/app/controllers/api/v1x0/requests_controller.rb
+++ b/app/controllers/api/v1x0/requests_controller.rb
@@ -4,6 +4,11 @@ module Api
       include Mixins::IndexMixin
       include Mixins::RBACMixin
 
+      PERSONA_HEADER    = 'x-rh-persona'.freeze
+      PERSONA_ADMIN     = 'approval/admin'.freeze
+      PERSONA_APPROVER  = 'approval/approver'.freeze
+      PERSONA_REQUESTER = 'approval/requester'.freeze
+
       before_action :read_access_check, :only => %i[show]
       before_action :create_access_check, :only => %i[create]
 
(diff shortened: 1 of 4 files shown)


spec/controllers/v1.0/requests_controller_spec.rb
```diff
@@ -4,6 +4,10 @@
   let(:encoded_user) { encoded_user_hash }
   let(:tenant) { create(:tenant) }
 
+  let(:headers_with_admin)     { default_headers.merge(described_class::PERSONA_HEADER => described_class::PERSONA_ADMIN) }
```
mkanoor: @bzwei This looks good @syncrou is trying to add regular users and admin users with a trait. Admin users would have the is_org_admin set so we might want to get the right user in there

mkanoor:
@bzwei You might also want to mention that setting the person doesn't bypass the RBAC check, we still do the RBAC check before we pass back the data

spec/controllers/v1.0/requests_controller_spec.rb
bzwei: our admin means Approver Administrator, not org_admin, nor rbac_admin.

app/controllers/api/v1x0/requests_controller.rb
```diff
@@ -4,6 +4,11 @@ class RequestsController < ApplicationController
       include Mixins::IndexMixin
       include Mixins::RBACMixin
 
+      PERSONA_HEADER    = 'x-rh-persona'.freeze
```
mkanoor: @bzwei this should be header key
PERSONA_HEADER_KEY
And should that go in here
https://github.com/ManageIQ/manageiq-api-common/blob/83aa89d7049de6e7864750bee980c04cdc5f9667/lib/manageiq/api/common/request.rb#L12

bzwei: https://github.com/ManageIQ/manageiq-api-common/pull/109 has been created for this purpose. Shall we wait until it is merged or make a follow-up PR to reference it?

bzwei:
> @bzwei You might also want to mention that setting the person doesn't bypass the RBAC check, we still do the RBAC check before we pass back the data

Done

PR #181, opened 2019-10-04
changed to return approver resource id list based on stages

hsong-rh:
In this PR, RBAC checking for approver on the resources is changed into `stage` level, rather than `request` as before. So approver can only access to the stage which either has been processed before, or is the current one and pending for process. 

https://projects.engineering.redhat.com/browse/SSP-789

### DIFF
--- a/app/controllers/api/v1x0/mixins/rbac_mixin.rb
+++ b/app/controllers/api/v1x0/mixins/rbac_mixin.rb
@@ -7,6 +7,8 @@ module Api
         ADMIN_ROLE = 'Approval Administrator'.freeze
         APPROVER_ROLE = 'Approval Approver'.freeze
 
+        APPROVER_ROLE_PREFIX = 'approval-group-'.freeze
+
         # create action only needs class level permission check
         def create_access_check
           permission_check('create')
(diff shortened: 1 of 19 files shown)


app/controllers/api/v1x0/mixins/rbac_mixin.rb
```diff
@@ -105,36 +130,50 @@ def approver_request_ids
...
+        def approver_stage_ids
+          group_ids = assigned_group_ids
+          Rails.logger.info("Groups from assigned roles: #{group_ids}")
+
+          # stage ids after filtering with user groups
+          stage_ids = stages_groups.select { |_stage_id, group_id| group_ids.include?(group_id) }.keys
+          Rails.logger.info("Approvable stage ids after filtering with assigned groups: #{stage_ids}")
+
+          # another filtering based on stage index inside request
+          stage_ids.select do |stage_id|
```
bzwei: select stages with state being either notified or finished.

app/controllers/api/v1x0/mixins/rbac_mixin.rb
```diff
@@ -105,36 +130,50 @@ def approver_request_ids
...
+            end
+
+            stage.index_of_request <= stage.request.active_stage_number
+          end
         end
 
-        # Action ids associated with request ids #{request_ids}
-        def action_ids(request_ids)
-          Action.where(:stage_id => stage_ids(request_ids)).pluck(:id).sort
+        def assigned_group_ids
```
bzwei: group_refs

lib/rbac/roles.rb
```diff
@@ -2,6 +2,8 @@ module RBAC
   class Roles
     attr_reader :roles
 
+    APPROVER_ROLE_PREFIX = 'approval-group-'.freeze
```
bzwei: Should the constant be added here? Didn't you move `lib/rbac` to `manageiq-api-common`?

PR #187, opened 2019-10-10
Schema, model, service, and test for TagLink

bzwei:


app/services/workflow_link_service.rb
```diff
@@ -0,0 +1,28 @@
...
+    nil
+  end
+
+  def unlink(options)
+    TagLink.find_by(tag_link(options)).try(:destroy)
+  end
+
+  def find(options_arr)
+    options_arr.collect do |options|
+      TagLink.where(tag_link(options)).pluck(:workflow_id).first
```
mkanoor: @bzwei Is this returning an array of workflow ids? How would we know which array entry corresponds to which workflow.

app/services/workflow_link_service.rb
```diff
@@ -0,0 +1,28 @@
...
+    end
+  end
+
+  private
+
+  def tag_link(options)
+    options.merge(:workflow_id => workflow_id)
```
mkanoor: @bzwei Why is this merging the workflow_id, shouldn't it be using the object_type, app_name to do the match and then pickup the TagLink object.

app/services/workflow_link_service.rb
```diff
@@ -0,0 +1,28 @@
+class WorkflowLinkService
+  attr_accessor :workflow_id
+
+  def initialize(workflow_id)
+    self.workflow_id = workflow_id
+  end
+
+  def link(options)
+    TagLink.where(tag_link(options)).first_or_create
+    nil
```
hsong-rh: Why we force to return `nil` here?

app/services/workflow_link_service.rb
```diff
@@ -0,0 +1,22 @@
+class WorkflowLinkService
+  attr_accessor :workflow_id
+
+  def initialize(workflow_id)
+    self.workflow_id = workflow_id
+  end
+
+  def link(options)
```
hsong-rh: Feel it's more straightforward to use `tags` than `options`, as parameter's name. Same thing for other methods.

app/services/workflow_link_service.rb
bzwei: This is like an array transform method many languages support: the output array has the same  length of the input, one to one mapping.

app/services/workflow_link_service.rb
bzwei: The caller only needs to know whether it succeeds or not. No `TagLink` object is required.

PR #189, opened 2019-10-11
Fix tag_link uniqueness validation

bzwei:
The uniqueness validation should take tenant_id into consideration. ActsAsTenant does not automatically add the scope to validation.


mkanoor:
@bzwei Can we get the Rubocop warning fixed

PR #192, opened 2019-10-15
Schema and model change to add sequence to workflow

bzwei:
Backend support to enable workflow sequencing management per tenant using ActsAsList.

When a new workflow is created, the sequence is automatically assigned to the next available value if it is not provided by the caller.

Workflow update API allows the caller to directly update. No dedicated API and service class are needed.

When a workflow sequence is changed, all affected workflows' sequences will be automatically updated.

### DIFF
--- a/spec/models/workflow_spec.rb
+++ b/spec/models/workflow_spec.rb
@@ -16,29 +16,25 @@ RSpec.describe Workflow, :type => :model do
     before { create_list(:workflow, 5) }
 
     it 'lists workflows in sequence ascending order' do
-      orders = Workflow.all.pluck(:sequence)
-      sorted_orders = orders.sort
-      expect(orders).to eq(sorted_orders)
+      orders = Workflow.pluck(:sequence)
+      expect(orders).to eq(orders.sort)
     end
 
     it 'places newly created workflow to the end of ascending list' do
-      old_last = Workflow.all.last
-      new_workflow = create(:workflow)
-      expect(new_workflow.sequence).to be > old_last.sequence
+      old_last = Workflow.last
+      expect(create(:workflow).sequence).to be > old_last.sequence
     end
 
     it 'moves up an workflow sequence' do
-      old_ids = Workflow.all.pluck(:id)
+      old_ids = Workflow.pluck(:id)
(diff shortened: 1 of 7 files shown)


app/services/workflow_sequence_service.rb
```diff
@@ -0,0 +1,15 @@
+class WorkflowSequenceService
+  def list_by_sequence
+    ActsAsTenant.current_tenant.workflows
+  end
+
+  def update_sequences(workflow_ids)
```
hsong-rh: This `workflow_ids` should contain all existing workflow ids, right? Partial of workflow ids will mess the `sequence`. Suggest to rename to `update_all_sequences`, if this is the case.

bzwei:
I have removed the service class as the design is now much simplified since the initial work.

spec/models/workflow_spec.rb
```diff
@@ -8,6 +8,40 @@
...
+      old_last = Workflow.all.last
+      new_workflow = create(:workflow)
+      expect(new_workflow.sequence).to be > old_last.sequence
+    end
+
+    it 'moves up an workflow sequence' do
+      old_ids = Workflow.all.pluck(:id)
+      Workflow.find(old_ids[3]).update(:sequence => Workflow.find(old_ids[1]).sequence)
+      new_ids = Workflow.all.pluck(:id)
+      expect(new_ids).to eq([old_ids[0], old_ids[3], old_ids[1], old_ids[2], old_ids[4]])
```
gmcculloug: Same, drop `new_ids` and just use `Workflow.pluck(:id)` in the `expect` call.

spec/models/workflow_spec.rb
```diff
@@ -8,6 +8,40 @@
...
+    end
+
+    it 'moves down an workflow sequence' do
+      old_ids = Workflow.all.pluck(:id)
+      Workflow.find(old_ids[1]).update(:sequence => Workflow.find(old_ids[3]).sequence)
+      new_ids = Workflow.all.pluck(:id)
+      expect(new_ids).to eq([old_ids[0], old_ids[2], old_ids[3], old_ids[1], old_ids[4]])
```
gmcculloug: Same here.

PR #191, opened 2019-10-15
Add tag link feature in openapi spec and controllers

hsong-rh:
Changed Openapi.json and workflow controller to enable following endpoints:
`workflows/{id}/link`,  `workflows/{id}/unlink`, `workflows/unlink` and `workflows/resolve`

PR #193, opened 2019-10-16
add workflow_unlink_service

bzwei:
Create a separate service class to do the unlinking. It does not delete the entry from `tag_links` table. Instead it should attempt to delete resolved tag name from the resource object (to be followed up by another PR).


bzwei:
for #191

PR #194, opened 2019-10-23
Request with children

bzwei:
We are in the process to replace stages with sub requests.
In this work a new column `parent_id` is added to `requests` table and `acts_as_tree` is used to add parent-children association.

Stages will be completely removed in a future PR.

(Only review the second commit)


app/models/request.rb
```diff
@@ -4,6 +4,7 @@ class Request < ApplicationRecord
   include OwnerField
 
   acts_as_tenant(:tenant)
+  acts_as_tree
```
gmcculloug: This call does not require any configuration?

app/controllers/api/v1x0/requests_controller.rb
```diff
@@ -31,6 +31,10 @@ def index
         collection(index_scope(reqs))
       end
 
+      def index_scope(relation)
+        super(relation.includes(:children))
```
mkanoor: @bzwei does this need RBAC?

app/models/request.rb
```diff
@@ -30,6 +31,14 @@ def current_stage
     stages.find_by(:state => [Stage::NOTIFIED_STATE, Stage::PENDING_STATE])
   end
 
+  def num_children
```
gmcculloug: Where does `num_children` get called from and we really should fully spell out the name.  `number_of_children` or `child_size`.

app/models/request.rb
```diff
@@ -30,6 +31,14 @@ def current_stage
...
+    children.size
+  end
+
+  def num_finished_children
```
gmcculloug: same here

app/models/request.rb
bzwei: The optional parameter is `:sort` and we already sort requests by creation time.

app/controllers/api/v1x0/requests_controller.rb
bzwei: `super` calls RBAC

app/models/request.rb
bzwei: They are replacing `total_stages` and `current_active_stage`. They are included in openapi spec for UI to quickly display the overall progress.

app/models/request.rb
gmcculloug: What if that current sort goes away?  The API is closer to supporting a `sort_by` option and we would likely remove the hardcoded value from the model and allow the UI to request the sort order.  The current default scope seems to supersedes any other  sort value passed so it limits what the API callers (like the UI) can do.

Does it make sense to define it if the default scope was not there?

bzwei: sorting is optional for `ActsAsTree`. If we depend on `sort_by` option from caller, we have more reason not  to add the `:sort` option in `acts_as_tree()`.

PR #200, opened 2019-10-31
Move content and context into a separate table and share with children

bzwei:
Move `content` and `context` into table `request_contexts`.
Share the same `request_context` when a child request is created.


mkanoor:
@bzwei can you fix the rubocop warnings

bzwei:
The only rubocop warning remained it the timestamp which I don't think is needed in this case because the `request_context` is never used standalone. It is always tied to a request.

PR #201, opened 2019-10-31
Added support for remote tagging

mkanoor:
Uses Faraday to post tags to remote objects


app/services/workflow_link_service.rb
```diff
@@ -1,25 +1,33 @@
...
-  # TODO: create tag name based on workflow id
-  def tag_name
-    "#{TAG_PREFIX}#{workflow_id}"
+  def fq_tag_name
+    "/#{TAG_NAMESPACE}/#{TAG_NAME}=#{workflow_id}"
+  end
+
+  def approval_tag
+    { :name      => TAG_NAME,
+      :value     => workflow_id.to_s,
```
bzwei: Will all remote services allow the same namespace and name to have multiple values?

app/services/workflow_link_service.rb
```diff
@@ -1,25 +1,33 @@
...
+  TAG_NAMESPACE = 'approval'.freeze
+  TAG_NAME      = 'workflows'.freeze
 
   def initialize(workflow_id)
     self.workflow_id = workflow_id
   end
 
   def link(tag_attrs)
     TagLink.find_or_create_by!(tag_link(tag_attrs))
+    RemoteTaggingService.new(tag_attrs).process('add', approval_tag)
```
bzwei: What happen if the tag already exists at the remote service? Will it return an error? Should we consume the error and continue?

app/services/remote_tagging_service.rb
```diff
@@ -0,0 +1,70 @@
+require 'faraday'
+class RemoteTaggingService
+  def initialize(options)
+    @app_name = options[:app_name]
+    @object_type = options[:object_type]
+    @object_id = options[:object_id]
+  end
+
+  def process(action, tag)
+    post_request(action_url(action), tag)
```
bzwei: Will you support deleting tag in the near future? Will it still use POST but not DELETE?

app/services/remote_tagging_service.rb
```diff
@@ -0,0 +1,70 @@
...
+    self
+  end
+
+
+  def self.remotes
+    [{ :app_name => 'topology', :object_type => 'ServiceInventory', :url => proc {topo_url} },
+     { :app_name => 'topology', :object_type => 'Credential',       :url => proc {topo_url} },
+     { :app_name => 'catalog',  :object_type => 'PortfolioItem',    :url => proc {catalog_url} },
+     { :app_name => 'catalog',  :object_type => 'Portfolio',        :url => proc {catalog_url} },
+     { :app_name => 'sources',  :object_type => 'Source',           :url => proc {sources_url} }]
```
bzwei: Let's consider using OpenShift ConfigMap to store the mapping from app/type -> url.
The key can even include the action, such as `add/topology/ServiceInventory`.

The url can contain an object_id placeholder. The in our code we only need to do a replacement to get a fully qualified URL.

app/services/workflow_link_service.rb
mkanoor: Yes. If the name and namespace remains the same but the value changes it will result in different tags being created. If you add the same tag multiple times with the same name, namespace and value it doesn't add the same thing multiple times.

app/services/workflow_link_service.rb
mkanoor: @bzwei If it exists we get back a 200 and nothing bad happens

app/services/remote_tagging_service.rb
mkanoor: @bzwei The add is a temporary solution till we can get the full create/delete tags from the common gem. The API definition might change so once that happens I will add the delete.

app/services/remote_tagging_service.rb
mkanoor: @bzwei Future

spec/services/workflow_link_service_spec.rb
```diff
@@ -1,10 +1,26 @@
 RSpec.describe WorkflowLinkService do
+  around do |example|
+    ManageIQ::API::Common::Request.with_request(default_request_hash) { example.call }
+  end
+
   let(:workflow) { create(:workflow, :with_tenant, :group_refs => [990]) }
-  let(:obj_a) { {:object_type => 'inventory', :app_name => 'topology', :object_id => 'abc'} }
+  let(:object_id) { '123' }
```
bzwei: what's the purpose to separate `object_id`?

spec/services/remote_tagging_service_spec.rb
```diff
@@ -0,0 +1,108 @@
...
+  let(:object_id) { '123' }
+  let(:options) { {:object_type => object_type, :app_name => app_name, :object_id => object_id} }
+  let(:approval_tag) do
+    { :namespace => WorkflowLinkService::TAG_NAMESPACE,
+      :name      => WorkflowLinkService::TAG_NAME,
+      :value     => "100" }
+  end
+  let(:http_status) { [200, 'Ok'] }
+  let(:headers)     do
+    { 'Content-Type' => 'application/json' }.merge(default_headers)
```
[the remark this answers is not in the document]
mkanoor: @bzwei Yes we can change but it would have to be done in a separate PR leads to other spec failures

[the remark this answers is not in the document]
mkanoor: @bzwei The default headers is used for both GET and POST calls and typically we dont set Content-Type in a GET call and only in a POST call, when we are sending the data. When we do a POST in catalog we have modified the specs to automatically add :as => :json which converts the hash to json and also sets the content type that code is missing in approval

https://github.com/ManageIQ/catalog-api/pull/369

PR #204, opened 2019-11-05
Remove stage model/service and associate actions to request

bzwei:
This PR address changes to models only.

Temporarily disable failed spec tests due to incomplete work on service classes and controllers.

It breaks APIs which will be fixed through follow-up PRs.

Update:
Stage controller is removed too.


bzwei:
@mkanoor can you review and merge?  I have another PR depends on it. Once both PRs are merged the tag based approval for service ordering can be completed and tested on CI.

PR #205, opened 2019-11-06
Use the request validator from common gem

hsong-rh:
Similar changes as https://github.com/ManageIQ/catalog-api/pull/402:
1. All 422 error codes are changed into 400;
2. Raise error if any parameters is missing;


app/controllers/api/v1x0/requests_controller.rb
```diff
@@ -32,7 +32,9 @@ def index_scope(relation)
       private
 
       def request_params
-        params.permit(:name, :description, :tag_resources => [], :content => {})
+        params.permit(:name, :description, :workflow_id, :requester_name,
```
mkanoor: @hsong-rh Whereever request_params is being used we can use 
https://github.com/ManageIQ/manageiq-api-common/blob/8e76ff7192ed56d84a64000dab7c9f0dc80534cc/lib/manageiq/api/common/application_controller_mixins/parameters.rb#L11
Since your controllers are using that mixing
An example is here
https://github.com/ManageIQ/catalog-api/blob/d2a85e62b47b060631554ba29762a2f71875966a/app/controllers/api/v1/portfolio_items_controller.rb#L28

hsong-rh: 👍 Will change accordingly after https://github.com/ManageIQ/manageiq-api-common/pull/133 is merged.

hsong-rh:
Depends on https://github.com/ManageIQ/manageiq-api-common/pull/133

app/controllers/api/v1x0/workflows_controller.rb
```diff
@@ -76,12 +77,8 @@ def rbac_scope(relation)
         relation
       end
 
-      def workflow_params
-        params.permit(:object_id, :object_type, :app_name, :name, :description, :group_refs => [])
-      end
-
       def attrs
-        workflow_params.slice(:object_id, :object_type, :app_name).to_unsafe_h
+        params.slice(:object_id, :object_type, :app_name).to_unsafe_h
```
mkanoor: @hsong-rh do we need this since the resource object only has those exact 3 attributes
https://github.com/ManageIQ/approval-api/blob/c2725503a8a2f8fa408c5253b82b1e2d6b5158f9/public/doc/openapi-3-v1.0.0.json#L1340
can't we just use params_for_create and it would do the seem thing and we won't need the attrs method

public/doc/openapi-3-v1.0.0.json
```diff
@@ -1110,7 +1110,7 @@
                 "description": "Query by id",
                 "required": true,
                 "schema": {
-                    "type": "string"
+                    "$ref": "#/components/parameters/Ident"
```
mkanoor: @hsong-rh can we call this Id or fully spell it out as Identity.
https://github.com/ManageIQ/topological_inventory-api/blob/daa581abc260a37fe08fb3f7bc254cb259e0a6a3/public/doc/openapi-3-v1.0.0.json#L8008

spec/controllers/v1.0/workflows_controller_spec.rb
```diff
@@ -329,7 +329,7 @@
       before do
         allow(rs_class).to receive(:paginate).and_return([])
         allow(roles_obj).to receive(:roles).and_return([admin_role])
-        patch "#{api_version}/workflows/#{id}", :params => valid_attributes, :headers => default_headers
+        patch "#{api_version}/workflows/#{id}", :params => valid_attributes, :headers => default_headers, :as => :json
```
mkanoor: @hsong-rh since this happens by default for specs with type => request do we need to specify the :as => :json since the common gem will do that for you

spec/controllers/v1.0/workflows_controller_spec.rb
```diff
@@ -482,35 +482,34 @@
...
+    let(:obj_a) { { :object_type => 'ServiceInventory', :app_name => 'topology', :object_id => '123'} }
+    let(:obj_b) { { :object_type => 'Portfolio', :app_name => 'catalog', :object_id => '123'} }
     before do
       allow(AddRemoteTags).to receive(:new).with(obj_a).and_return(add_tag_svc)
       allow(add_tag_svc).to receive(:process).with(tag).and_return(add_tag_svc)
       allow(GetRemoteTags).to receive(:new).with(obj_a).and_return(get_tag_svc)
       allow(GetRemoteTags).to receive(:new).with(obj_b).and_return(get_tag_svc)
       allow(get_tag_svc).to receive(:process).and_return(get_tag_svc)
-      post "#{api_version}/workflows/#{id}/link", :params => obj_a, :headers => default_headers
+      post "#{api_version}/workflows/#{id}/link", :params => obj_a, :headers => default_headers, :as => :json
```
mkanoor: @hsong-rh :as => :json should not be needed if the spec has the proper type the common gem should do this for you

spec/controllers/v1.0/workflows_controller_spec.rb
mkanoor: @hsong-rh Maybe https://github.com/ManageIQ/catalog-api/blob/a63a7704f70cb32f78b0bac1974be1038262f6d8/spec/spec_helper.rb#L27 is missing in your spec_helper.rb

PR #206, opened 2019-11-07
Fetch sorted workflows based on TagResources

mkanoor:
Catalog would be sending a collection of TagResources and this
code will return a list of workflows sorted by the sequence number

### DIFF
--- a/app/services/workflow_find_service.rb
+++ b/app/services/workflow_find_service.rb
@@ -16,7 +16,7 @@ class WorkflowFindService
       tag_names = tr['tags'].map { |tag| "/#{tag['namespace']}/#{tag['name']}=#{tag['value']}" }
       params = {:app_name => tr['app_name'], :object_type => tr['object_type'], :tag_name => tag_names}
       query =
-        if i == 0
+        if i.zero?
           query.where(params)
         else
           query.or(TagLink.where(params))
--- a/app/services/workflow_find_service.rb
+++ b/app/services/workflow_find_service.rb
@@ -22,7 +22,7 @@ class WorkflowFindService
           query.or(TagLink.where(params))
         end
     end
-    Workflow.where(:id => query).order(:sequence)
+    Workflow.where(:id => query)
   end
 
   def fq_tag_names(tag_attrs)
--- a/app/services/workflow_find_service.rb
+++ b/app/services/workflow_find_service.rb
@@ -8,8 +8,21 @@ class WorkflowFindService
   end
 
   # find workflows from a collection of [app_name, object_type, [namespace, key, value]]
-  def find_by_tag_resources(_tag_resources)
-    []
+  def find_by_tag_resources(tag_resources)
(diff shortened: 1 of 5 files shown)


mkanoor:
The travis failure is happening because the strong parameter check is dropping the tag_resource parameters since it's an array of hashes.
https://github.com/ManageIQ/approval-api/blob/c2725503a8a2f8fa408c5253b82b1e2d6b5158f9/app/controllers/api/v1x0/requests_controller.rb#L35

mkanoor:
@bzwei please review

app/services/workflow_find_service.rb
```diff
@@ -8,8 +8,21 @@ def find(tag_attrs)
...
+      tag_names = tr['tags'].map { |tag| "/#{tag['namespace']}/#{tag['name']}=#{tag['value']}" }
+      params = {:app_name => tr['app_name'], :object_type => tr['object_type'], :tag_name => tag_names}
+      query =
+        if i.zero?
+          query.where(params)
+        else
+          query.or(TagLink.where(params))
+        end
+    end
+    Workflow.where(:id => query).order(:sequence)
```
bzwei: No need to order by sequence. The default scope already has it.

spec/services/workflow_find_service_spec.rb
```diff
@@ -30,4 +30,23 @@
       expect(another_workflows).to eq([])
     end
   end
+
+  describe 'find_by_tag_resources' do
```
bzwei: describe '#find_by_tag_resources'

spec/services/workflow_find_service_spec.rb
```diff
@@ -30,4 +30,23 @@
...
+    include_context "tag_resource_objects"
+
+    it 'finds sorted workflow based on tags' do
+      workflows = subject.find_by_tag_resources([tag_resource1])
+      expect(workflows).to eq([workflow2, workflow1])
+    end
+
+    it 'if tags is empty returns empty array' do
```
bzwei: better reads `it 'returns an empty array if tags are empty'`

spec/services/workflow_find_service_spec.rb
```diff
@@ -30,4 +30,23 @@
...
+      workflows = subject.find_by_tag_resources([tagless_resource])
+      expect(workflows).to be_empty
+    end
+
+    it 'if input is empty array return empty array' do
```
bzwei: better reads `it 'returns an empty array if input is empty'`

spec/services/workflow_find_service_spec.rb
```diff
@@ -30,4 +30,23 @@
...
+      workflows = subject.find_by_tag_resources([])
+      expect(workflows).to be_empty
+    end
+  end
```
bzwei: can we have a test `it 'returns an empty array if tags find no links`?

PR #209, opened 2019-11-12
Added custom exception mappings

hsong-rh:
Similar work as https://github.com/ManageIQ/catalog-api/pull/472, add exception mappings to replace error code 422 with 400.


app/controllers/concerns/exception_handler.rb
```diff
@@ -7,24 +7,12 @@ module ExceptionHandler
...
-    rescue_from ActiveRecord::RecordInvalid do |e|
-      json_response({ :message => e.message }, :unprocessable_entity)
-    end
-
-    rescue_from ActionController::ParameterMissing do |e|
-      json_response({ :message => e.message }, :unprocessable_entity)
-    end
-
-    rescue_from ManageIQ::API::Common::EntitlementError, Exceptions::NotAuthorizedError do |e|
+    rescue_from ManageIQ::API::Common::EntitlementError do |e|
```
mkanoor: @hsong-rh Can't we get rid of all the rescue_froms and have the common gem handle it. And we can add the exceptions to custom_exception_mappings. The common gem can handle nested exceptions, since each exception has a cause attribute which allows us to link exceptions together.

hsong-rh: @mkanoor I changed to use manageiq-api-common 2.1 to handle all exceptions.

PR #210, opened 2019-11-12
Used shared open api serializer

hsong-rh:


public/doc/openapi-3-v1.0.0.json
```diff
@@ -1296,6 +1296,12 @@
                     "description": {
                         "type": "string",
                         "readOnly": true
+                    },
+                    "created_at": {
```
bzwei: Although adding this field does not hurt, why it is mandatory?

hsong-rh: Because in spec file, we expect `created_at` in `iso.8601` format. We have to add here so the out resource has this attribute.

bzwei: we should change spec test file rather than api spec.

PR #211, opened 2019-11-13
Fixed client generating errors in open api spec

hsong-rh:


bzwei:
@hsong-rh can you explain why we need to switch back to string? I thought the use of Identity is required by the validation requirement from api-common.

hsong-rh:
@bzwei It's not switch back, in `identity` it has been defined as string in a particular pattern. This pattern is required by api-common. The reason we defined it as `identity` is willing to share it among other parameters. Due to the caused syntax errors, here just changed back to define it directly in the schema.

PR #212, opened 2019-11-13
Changed resolve/link/unlink into GET/POST actions

hsong-rh:
Due to the request validation, here we changed:
1. replace ` POST /workflows/resolve` with `GET /workflows?remote_object_query_params` to resolve workflows based on remote object;
2. remove `POST /workflows/unlink` with `DELETE /workflows/unlink`

### DIFF
--- a/app/controllers/api/v1x0/mixins/index_mixin.rb
+++ b/app/controllers/api/v1x0/mixins/index_mixin.rb
@@ -15,8 +15,8 @@ module Api
           resp = ManageIQ::API::Common::PaginatedResponse.new(
             :base_query => filtered(scoped(base_query)),
             :request    => request,
-            :limit      => params[:limit],
-            :offset     => params[:offset]
+            :limit      => pagination_limit,
+            :offset     => pagination_offset
           ).response
 
           json_response(resp)
--- a/app/controllers/api/v1x0/mixins/index_mixin.rb
+++ b/app/controllers/api/v1x0/mixins/index_mixin.rb
@@ -15,8 +15,8 @@ module Api
           resp = ManageIQ::API::Common::PaginatedResponse.new(
             :base_query => filtered(scoped(base_query)),
             :request    => request,
-            :limit      => pagination_limit,
-            :offset     => pagination_offset
+            :limit      => params[:limit],
+            :offset     => params[:offset]
           ).response
 
           json_response(resp)
--- a/app/controllers/api/v1x0/workflows_controller.rb
+++ b/app/controllers/api/v1x0/workflows_controller.rb
@@ -24,6 +24,8 @@ module Api
         relation = if params[:template_id]
                      template = Template.find(params.require(:template_id))
                      template.workflows
+                   elsif resolvable?
+                     WorkflowFindService.new.find(resolve_params)
                    else
                      Workflow.all
                    end
@@ -50,14 +52,9 @@ module Api
       end
 
       def unlink
-        WorkflowUnlinkService.new(params[:id]).unlink(params_for_create.to_unsafe_h)
+        WorkflowUnlinkService.new(params[:id]).unlink(resolve_params)
 
         head :no_content
-      end
-
-      def resolve
-        found_workflows = WorkflowFindService.new.find(params_for_create.to_unsafe_h)
-        found_workflows.empty? ? head(:no_content) : json_response(found_workflows, :ok)
       end
 
       def update
@@ -75,6 +72,16 @@ module Api
         raise Exceptions::NotAuthorizedError, "Not Authorized for #{relation.model.table_name}" unless admin?
 
         relation
+      end
+
+      def resolve_params
+        params.slice(:object_type, :object_id, :app_name).to_unsafe_h
+      end
+
+      def resolvable?
+        raise Exceptions::UserError, "Invalid resolve params: #{resolve_params}" unless resolve_params.length.zero? || resolve_params.length == 3
+
+        !!(params[:app_name] && params[:object_id] && params[:object_type])
       end
     end
   end
--- a/app/controllers/api/v1x0/workflows_controller.rb
+++ b/app/controllers/api/v1x0/workflows_controller.rb
@@ -24,8 +24,8 @@ module Api
         relation = if params[:template_id]
                      template = Template.find(params.require(:template_id))
                      template.workflows
-                   elsif resolvable?
-                     WorkflowFindService.new.find(resolve_params)
+                   elsif for_resource_object?
+                     WorkflowFindService.new.find(resource_object_params)
                    else
                      Workflow.all
                    end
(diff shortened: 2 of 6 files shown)


bzwei:
@hsong-rh `DELETE` is not appropriate here. Let's stick with `POST` and remove `POST /workflows/unlink`. We don't expect a remote object to have too many tag links, so unlink them one by one should be OK if the user needs to unlink all workflows.

public/doc/openapi-3-v1.0.0.json
```diff
@@ -456,6 +456,15 @@
                 "description": "Return all approval workflows in ascending sequence order",
                 "operationId": "listWorkflows",
                 "parameters": [
+                    {
```
bzwei: Shouldn't them be grouped under `remote_object_query_params` as stated in the PR description? Let's use the name `resource_object` instead to align with what is in the spec already.

Best if could reuse `ResourceObject` schema.

app/controllers/api/v1x0/workflows_controller.rb
```diff
@@ -76,6 +73,16 @@ def rbac_scope(relation)
 
         relation
       end
+
+      def resolve_params
```
bzwei: rename to `resource_object_params`

app/controllers/api/v1x0/workflows_controller.rb
```diff
@@ -76,6 +73,16 @@ def rbac_scope(relation)
...
+        params.slice(:object_type, :object_id, :app_name).to_unsafe_h
+      end
+
+      def resolvable?
```
bzwei: rename to `for_resource_object?`

spec/controllers/v1.0/workflows_controller_spec.rb
```diff
@@ -501,16 +505,23 @@
...
+      get "#{api_version}/workflows", :params => obj_b, :headers => default_headers
 
-      expect(response).to have_http_status(204)
+      expect(response).to have_http_status(200)
+    end
+
+    it 'raises an user error' do
+      get "#{api_version}/workflows", :params => obj_c, :headers => default_headers
+
+      expect(response.body).to include("Exceptions::UserError: Invalid resolve params")
```
mkanoor: @hsong-rh can we check the errors instead something similar to this
https://github.com/ManageIQ/catalog-api/blob/08cccd4ff0f728bb83c0d3dc02074cafb0da1d1e/spec/requests/approval_request_spec.rb#L25

https://github.com/ManageIQ/catalog-api/blob/08cccd4ff0f728bb83c0d3dc02074cafb0da1d1e/spec/support/requests_spec_helper.rb#L19

app/controllers/api/v1x0/workflows_controller.rb
```diff
@@ -76,6 +73,16 @@ def rbac_scope(relation)
 
         relation
       end
+
+      def resource_object_params
+        params.slice(:object_type, :object_id, :app_name).to_unsafe_h
```
mkanoor: @hsong-rh do you want to save this using

@resource_object_params ||= params.slice(:object_type, :object_id, :app_name).to_unsafe_h

Because in the for_resource_object? you this method is going to get called twice

app/controllers/api/v1x0/mixins/index_mixin.rb
```diff
@@ -15,8 +15,8 @@ def collection(base_query)
           resp = ManageIQ::API::Common::PaginatedResponse.new(
             :base_query => filtered(scoped(base_query)),
             :request    => request,
-            :limit      => pagination_limit,
-            :offset     => pagination_offset
+            :limit      => params[:limit],
```
bzwei: Make no change to the mixin which applies to all classes. Override the `collection` method only in the `Workflow` class.

app/controllers/api/v1x0/workflows_controller.rb
```diff
@@ -76,6 +73,27 @@ def rbac_scope(relation)
...
+          :request    => request,
+          :limit      => params[:limit],
+          :offset     => params[:offset]
+        ).response
+
+        json_response(resp)
+      end
+
+      def resource_object_params
+        @resource_object_params ||= params.slice(:object_type, :object_id, :app_name).to_unsafe_h
```
bzwei: where is `resource_object_query_params` used? shouldn't it be `params[:resource_object_query_params].slice()`? same in the spec test.

public/doc/openapi-3-v1.0.0.json
```diff
@@ -978,6 +901,15 @@
                     "type": "object"
                 }
             },
+            "resource_object": {
+                "in": "query",
+                "name": "resource_object",
```
mkanoor: @hsong-rh Is this a nested query parameter?
Shouldn't it be 3 different parameters one for app_name, object_type, object_id?

hsong-rh: @mkanoor No, I just changed it into nested parameters, as Bill suggested.

mkanoor: @hsong-rh @bzwei 
Have you seen any examples of creating nested parameters, in the URL format you can only pass non nested separate query parameters. I haven't seen ways to group parameters together. We have to ensure that this would work thru curl.

bzwei: I wanted @hsong-rh to follow the example of the filter parameter.

mkanoor: @bzwei By keeping 3 separate query parameters we can have the Approval do the following
- Only app_name is specified List all workflows associated with the the app_name (micro service)
- When app_name and object_type is specified List all workflows related to the app_name and object type (all workflows for Topology and ServiceInventory)
- When app_name, object_type and object_id are specified list all workflows related to a specific object from a specific service. This requires us to go out to the remote service and collect the tags and then filter out the associated workflows.

bzwei: @mkanoor Do we have a use case that requires mentioned queries? Only the third one is implemented in the PR.
If the three attributes are requests' native columns what you suggested is legitimate, but they belong to tag_links. We'd better to have a dedicated key to indicate they are not build-in attributes like the filter parameter does. Otherwise how do we distinguish if in the future we need to add a column called `object_type` or `object_id`?

mkanoor: @bzwei Either way we need to validate this change with a generated client code to ensure that it can serialize the nested query parameter. So lets generate a Approval Client and test the nested parameter and see how it escapes it.

hsong-rh: @mkanoor @bzwei 
I tested with both Curl and Postman. The query parameters in our case has to be defined in body for GET action. Based on our discussion, I will change back to use three separate query parameters, instead of current single `resource_object`

public/doc/openapi-3-v1.0.0.json
```diff
@@ -456,6 +456,15 @@
                 "description": "Return all approval workflows in ascending sequence order",
```
bzwei: We need to enhance the description to explain the use of resource object for filtering, and possibly with an example.

PR #213, opened 2019-11-18
Report 403 errors from HTTP Post calls

mkanoor:
The 403 from a direct HTTP call should be converted back to a well known exception Exceptions::NotAuthorizedError so we can forward it to the UI

### DIFF
--- a/app/services/remote_tagging_service.rb
+++ b/app/services/remote_tagging_service.rb
@@ -1,6 +1,6 @@
 require 'faraday'
 class RemoteTaggingService
-  VALID_200_CODES = [200, 201, 202, 204]
+  VALID_200_CODES = [200, 201, 202, 204].freeze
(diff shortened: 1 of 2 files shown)


app/services/remote_tagging_service.rb
```diff
@@ -47,25 +47,32 @@ def service_url
...
-
+    check_for_exceptions(response, "Error getting tags")
     response
   end
 
+  def check_for_exceptions(response, message_prefix)
+    if response.status == 403
+      raise Exceptions::NotAuthorizedError, response.reason_phrase
+    else
+      raise "#{message_prefix} #{response.reason_phrase}" unless response.status == 200
```
bzwei: shall we accept all 200 return codes, for example 204?

mkanoor:
@bzwei Please review

PR #215, opened 2019-11-20
Changed to use shared RBAC service from api common

hsong-rh:


app/controllers/api/v1x0/mixins/rbac_mixin.rb
```diff
@@ -151,8 +151,8 @@ def owner_stage_ids(request_ids)
 
         # The accessible workflow ids for approver
         def workflow_ids
-          approval_access = RBAC::Access.new('workflows', 'approve').process
-          approval_access.send(:generate_ids)
+          approval_access = Insights::API::Common::RBAC::Access.new('workflows', 'approve').process
+          approval_access.send(:ids)
```
bzwei: should not have called this private method

hsong-rh: Currently the method `ids` is defined as `private` in insights-api-common. We can change it later to allow us to call it directly here.

bzwei: `id_list` method is sufficient. It calls `ids` in the backend.

hsong-rh: Just realized this line is extra and should remove it.

PR #216, opened 2019-11-21
Execute a request through sub_requests

bzwei:
Rewrite the login to create and execute a request. 
Create sub_requests based on resolved workflows. 
Execute a request through actions.
Update the root request based on its children's states.
Remove acts_as_tree gem since our use of child_parent relation is very simple.
Disable some spec tests - to be added with follow-up PRs.

Depends on #204

### DIFF
--- a/app/models/request.rb
+++ b/app/models/request.rb
@@ -4,11 +4,13 @@ class Request < ApplicationRecord
   include OwnerField
 
   acts_as_tenant(:tenant)
-  acts_as_tree
 
   belongs_to :request_context, :optional => false
   belongs_to :workflow
   has_many :actions, -> { order(:id => :asc) }, :dependent => :destroy, :inverse_of => :request
+
+  belongs_to :parent,   :foreign_key => :parent_id, :class_name => 'Request', :inverse_of => :children
+  has_many   :children, :foreign_key => :parent_id, :class_name => 'Request', :inverse_of => :parent, :dependent => :destroy
 
   validates :name,     :presence  => true
   validates :state,    :inclusion => { :in => STATES }
@@ -31,6 +33,8 @@ class Request < ApplicationRecord
   end
 
   def invalidate_number_of_finished_children
+    return if number_of_children.zero?
+
     update(:number_of_finished_children => children.count { |child| Request::FINISHED_STATES.include?(child.state) })
   end
 
@@ -38,6 +42,26 @@ class Request < ApplicationRecord
     self.class.create!(:name => name, :description => description, :owner => owner, :requester_name => requester_name, :parent_id => id, :request_context_id => request_context_id).tap do
       invalidate_number_of_children
     end
+  end
+
+  def root
+    root? ? self : parent
+  end
+
+  def root?
+    parent_id.nil?
+  end
+
+  def leaf?
+    number_of_children.zero?
+  end
+
+  def pure_leaf?
+    !root? && leaf?
+  end
+
+  def pure_root?
+    root? && !leaf?
   end
 
   def group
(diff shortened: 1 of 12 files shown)


app/services/request_create_service.rb
```diff
@@ -39,59 +64,45 @@ def auto_approve?
...
   end
 
   def auto_approve(request)
     sleep_time = ENV['AUTO_APPROVAL_INTERVAL'].to_f
 
-    start_first_stage(request)
-    request.stages.each { |stage| group_auto_approve(stage, sleep_time) }
+    start_request(request)
+
+    sub_requests = request.children
```
mkanoor: @bzwei 
If request.children is not a computed field we could use this.
```
sub_requests = request.children.empty? ? [request] : request.children
```

app/services/request_create_service.rb
```diff
@@ -39,59 +64,45 @@ def auto_approve?
...
-    request_started(request)
-    ActionCreateService.new(request.stages.first.id).create(
-      :operation    => Action::NOTIFY_OPERATION,
-      :processed_by => 'system'
-    )
-  end
+  def start_request(request)
+    sub_request_ids =
+      if request.leaf?
+        [request]
```
mkanoor: @bzwei should this be request.id since its collecting the sub_request_ids.

app/models/request.rb
```diff
@@ -40,6 +44,26 @@ def create_child
...
+
+  def root?
+    parent_id.nil?
+  end
+
+  def leaf?
+    number_of_children.zero?
+  end
+
+  def pure_leaf?
```
mkanoor: @bzwei Should this be called a branch or a subtree? leaf and pure_leaf are confusing.

app/services/request_update_service.rb
```diff
@@ -6,16 +6,139 @@ def initialize(request_id)
...
+
+  def notified(options)
+    request.update!(options.merge(:notified_at => DateTime.now))
+
+    EventService.new(request).approver_group_notified if request.leaf?
+
+    update_parent(options) if request.pure_leaf?
+  end
+
+  def completed(options)
```
mkanoor: @bzwei Can this be broken into smaller functions.

app/services/request_update_service.rb
```diff
@@ -6,16 +6,139 @@ def initialize(request_id)
...
+    finish_request(options[:decision]) if request.leaf?
+
+    return if request.state == Request::CANCELED_STATE
+
+    EventService.new(request).approver_group_finished if request.leaf?
+
+    if request.pure_leaf?
```
mkanoor: @bzwei This could be branch_completed/sub_tree_completed

app/services/request_update_service.rb
```diff
@@ -6,16 +6,139 @@ def initialize(request_id)
...
+      next unless leaf.state == Request::PENDING_STATE
+
+      ActionCreateService.new(leaf.id).create(:operation => Action::SKIP_OPERATION, :processed_by => 'system')
+    end
+  end
+
+  def peers_approved?(request)
+    peers = Request.where(:workflow_id => request.workflow_id, :parent_id => request.parent_id)
+
+    peers.each do |peer|
```
mkanoor: @bzwei 
Could this be
```
peers.any? { |peer| peer.decision != Request::APPROVED_STATUS } ? false : true
```

app/services/request_create_service.rb
```diff
@@ -39,59 +64,44 @@ def auto_approve?
...
-
-  def start_first_stage(request)
-    request_started(request)
-    ActionCreateService.new(request.stages.first.id).create(
-      :operation    => Action::NOTIFY_OPERATION,
-      :processed_by => 'system'
-    )
-  end
+  def start_request(request)
+    sub_request_ids =
```
mkanoor: @bzwei can this be made into a simple method called sub_request_ids and then you can directly use it line 100

PR #219, opened 2019-11-22
Reenable and fix request_create_service spec

bzwei:
Reenable and fix request_create_service spec

Fix bugs found through spec tests.

### DIFF
--- a/app/services/workflow_find_service.rb
+++ b/app/services/workflow_find_service.rb
@@ -9,7 +9,7 @@ class WorkflowFindService
 
   # find workflows from a collection of [app_name, object_type, [namespace, key, value]]
   def find_by_tag_resources(tag_resources)
-    return [] if tag_resources.empty?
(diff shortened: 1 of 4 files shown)


app/services/workflow_find_service.rb
```diff
@@ -9,7 +9,7 @@ def find(tag_attrs)
 
   # find workflows from a collection of [app_name, object_type, [namespace, key, value]]
   def find_by_tag_resources(tag_resources)
-    return [] if tag_resources.empty?
+    return [] if tag_resources.blank?
```
mkanoor: @bzwei is this because the caller is passing in a nil? Wouldn't that be caught by the openapi_parser validators

bzwei: yes for api calls. But it is also called by `RequestCreateService`. In a spec test I did not provide tag_resources which blowed up. I think this a safer change.

PR #222, opened 2019-11-26
Added /requests/:id/requests endpoint

hsong-rh:


app/controllers/api/v1/requests_controller.rb
```diff
@@ -22,11 +22,13 @@ def show
       end
 
       def index
-        collection(index_scope(Request.all))
-      end
+        requests = if params[:request_id]
+                     Request.find(params[:request_id])
```
bzwei: `.childen`

app/controllers/api/v1/requests_controller.rb
```diff
@@ -22,11 +22,13 @@ def show
...
+                   else
+                     Request.all
```
bzwei: `Request.where(:parent_id => nil)`. Only the root requests

bzwei:
spec tests will be done in a follow-up PR

PR #224, opened 2019-11-26
Set random_access_key when request starts and clear it after it completes

bzwei:


spec/services/request_create_service_spec.rb
```diff
@@ -100,7 +101,7 @@
           :reason         => 'System approved',
 
           :number_of_children          => 0,
-          :number_of_finished_children => 0
+          :number_of_finished_children => 0,
```
hsong-rh: Why do we need this change? typo?

PR #226, opened 2019-12-04
Cleanup generated ACLs when their associated workflow is removed

hsong-rh:
Add `WorkflowDeleteService` to cleanup ACLs when workflow is destroyed.

### DIFF
--- a/app/controllers/api/v1/workflows_controller.rb
+++ b/app/controllers/api/v1/workflows_controller.rb
@@ -34,14 +34,11 @@ module Api
       end
 
       def destroy
-        workflow = Workflow.find(params.require(:id))
-        workflow.destroy!
+        WorkflowDeleteService.new(params.require(:id)).delete
         head :no_content
       rescue ActiveRecord::InvalidForeignKey => e
         json_response({ :message => e.message }, :forbidden)
-      rescue ActiveRecord::RecordNotDestroyed
-        raise unless workflow.errors[:base].include?(Workflow::MSG_PROTECTED_RECORD)
-
+      rescue Exceptions::NotAuthorizedError
         json_response({ :message => Workflow::MSG_PROTECTED_RECORD }, :forbidden)
       end
 
--- a/app/services/workflow_delete_service.rb
+++ b/app/services/workflow_delete_service.rb
@@ -1,0 +1,30 @@
+class WorkflowDeleteService
+  attr_accessor :workflow
+
+  def initialize(workflow_id)
+    self.workflow = Workflow.find(workflow_id)
+  end
+
+  def delete
+    group_refs = workflow.group_refs
+
+    if group_refs.any?
+      aps = AccessProcessService.new
+
+      ContextService.new(Insights::API::Common::Request.current.to_h.transform_keys(&:to_s)).as_org_admin do
+        aps.remove_resource_from_groups(workflow.id, group_refs)
+      end
+    end
+
+    # TODO: remove remote tags
+    ##  tags = workflow.tag_links
+
+    ##  if tags.any?
+    ##    # DeleteRemoteTags.new(tags).process
+    ##  end
+
+    workflow.destroy!
+  rescue ActiveRecord::RecordNotDestroyed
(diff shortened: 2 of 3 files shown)


app/services/workflow_delete_service.rb
```diff
@@ -0,0 +1,23 @@
+class WorkflowDeleteService
+  attr_accessor :workflow
+
+  def initialize(workflow_id)
+    self.workflow = Workflow.find(workflow_id)
+  end
+
+  def cleanup
```
bzwei: just called `delete`

bzwei: also need to delete the entries in tag links table, but we also need to delete the actual tag remotely.

spec/services/workflow_delete_service_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  subject { described_class.new(workflow.id) }
+
+  context 'when update' do
+    before do
+      allow(AccessProcessService).to receive(:new).and_return(aps)
+      allow(aps).to receive(:remove_resource_from_groups)
+    end
+
+    it 'returns status code 204' do
```
bzwei: the description is wrong

spec/services/workflow_delete_service_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+      Insights::API::Common::Request.with_request(RequestSpecHelper.default_request_hash) do
+        subject.cleanup
+
+        expect(Workflow.find_by(:id => workflow.id)).to be_nil
```
bzwei: also need to assert the acls being removed, and tag links removed.

app/services/workflow_delete_service.rb
hsong-rh: Name is changed into `delete`. The local tag links will be cleaned up automatically. Remote tags need to implement `DeleteRemoteTags` service first. Marked `TODO` here, will add it later.

spec/services/workflow_delete_service_spec.rb
hsong-rh: `acls` is stub mockup. Tests for tag links will add later.

spec/services/workflow_delete_service_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  subject { described_class.new(workflow.id) }
+
+  context 'when delete' do
+    before do
+      allow(AccessProcessService).to receive(:new).and_return(aps)
+      allow(aps).to receive(:remove_resource_from_groups)
+    end
+
+    it 'returns nil after deletion' do
```
bzwei: `it 'deletes the workflow'`

app/controllers/api/v1/workflows_controller.rb
```diff
@@ -34,14 +34,11 @@ def index
...
-        workflow = Workflow.find(params.require(:id))
-        workflow.destroy!
+        WorkflowDeleteService.new(params.require(:id)).delete
         head :no_content
       rescue ActiveRecord::InvalidForeignKey => e
         json_response({ :message => e.message }, :forbidden)
-      rescue ActiveRecord::RecordNotDestroyed
-        raise unless workflow.errors[:base].include?(Workflow::MSG_PROTECTED_RECORD)
-
+      rescue Exceptions::NotAuthorizedError
```
bzwei: why to remove `RecordNotDestoryed` and why specially for `NotAuthorizedError`? Shouldn't it be commonly handled for all controller methods?

app/controllers/api/v1/workflows_controller.rb
```diff
@@ -34,14 +34,11 @@ def index
       end
 
       def destroy
-        workflow = Workflow.find(params.require(:id))
-        workflow.destroy!
+        WorkflowDeleteService.new(params.require(:id)).delete
         head :no_content
       rescue ActiveRecord::InvalidForeignKey => e
         json_response({ :message => e.message }, :forbidden)
```
bzwei: @lgalis suggests to improve the error message when the workflow still has requests link to it.
Maybe we should explicitly detect the association and not rely on the rails build-in protection. The logic should be placed in `WorkflowDeleteService`.

bzwei:
close in favor of #243

PR #228, opened 2019-12-05
Fix bypassing jbpm logic

bzwei:
https://projects.engineering.redhat.com/browse/SSP-1000

### DIFF
--- a/app/services/request_update_service.rb
+++ b/app/services/request_update_service.rb
@@ -125,7 +125,7 @@ class RequestUpdateService
 
   # start the external approval process if configured
   def start_request
-    return unless !bypass? || request.workflow.try(:external_processing?)
+    return unless !bypass? && request.workflow.try(:external_processing?)
 
     template = request.workflow.template
     processor_class = "#{template.process_setting['processor_type']}_process_service".classify.constantize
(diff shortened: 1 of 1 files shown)


app/services/request_update_service.rb
```diff
@@ -125,7 +125,7 @@ def update_parent(options)
 
   # start the external approval process if configured
   def start_request
-    return unless !bypass? || request.workflow.try(:external_processing?)
+    return unless !bypass? && request.workflow.try(:external_processing?)
```
gmcculloug: @bzwei @hsong-rh Complex `unless` conditions are hard to read and should be avoided.  I would strongly suggest changing this into an `if-condition`.

Note: You are using `unless` several times in this file and one of them includes the same complex condition being modified here.  Does that need to be modified as well?  See https://github.com/RedHatInsights/approval-api/blob/master/app/services/request_update_service.rb#L144

bzwei: I did not change to `if` because the part `!bypass? &&` lives only for a short time, shall be removed very soon.

hsong-rh: When BPM is reenabled, this `bypass` will be removed.

PR #233, opened 2019-12-11
Fixed to list child requests

hsong-rh:
For `owner` role,  current logic will filter out all sub requests by `:parent_id => nil`. Removed the filter to fix the problem.

### DIFF
--- a/spec/controllers/v1.0/requests_controller_spec.rb
+++ b/spec/controllers/v1.0/requests_controller_spec.rb
@@ -424,6 +424,30 @@ RSpec.describe Api::V1x0::RequestsController, :type => :request do
     end
   end
 
+  # Test suite for GET /requests/:request_id/requests
+  describe 'GET /requests/:request_id/requests' do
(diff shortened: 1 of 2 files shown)


spec/controllers/v1.0/requests_controller_spec.rb
```diff
@@ -424,6 +424,31 @@
     end
   end
 
+  # Test suite for GET /requests/:request_id/requests
+  describe 'GET /requests/:request_id/requests' do
+    let(:parent_request) { create(:request, :name => "parent", :owner => "jdoe", :number_of_children => 2, :number_of_finished_children => 0, :tenant_id => tenant.id) }
+    let!(:child_request_a) { create(:request, :owner => "jdoe", :parent_id => parent_request.id, :name => "child a", :workflow_id => workflow.id, :tenant_id => tenant.id) }
+    let!(:child_request_b) { create(:request, :owner => "jdoe", :parent_id => parent_request.id, :name => "child b", :workflow_id => workflow.id, :tenant_id => tenant.id) }
+    let(:request_id) { parent_request.id }
+    let(:group) { double }
```
bzwei: `double(:group, :name => 'foo')`, then you don't need Ln438

PR #234, opened 2019-12-12
Add endpoint /requests/:id/content

bzwei:
The return body will be an object (Hash).

Also expose `created_at` in `Request` schema.


bzwei:
cc @lgalis

bzwei:
@mkanoor please review and merge

PR #235, opened 2019-12-13
Support for new tagging changes

mkanoor:
Uses the latest tagging changes 
* tag end point to create and assign new tags
* untag end point to delete tags

Waiting on some of the other PRs to get merged first

https://github.com/RedHatInsights/catalog-api/pull/495

### DIFF
--- a/app/services/add_remote_tags.rb
+++ b/app/services/add_remote_tags.rb
@@ -1,6 +1,6 @@
 class AddRemoteTags < RemoteTaggingService
   def process(tag)
-    post_request(object_url, tag)
+    post_request(object_url, [tag])
     self
   end
 
--- a/app/services/add_remote_tags.rb
+++ b/app/services/add_remote_tags.rb
@@ -3,4 +3,8 @@ class AddRemoteTags < RemoteTaggingService
     post_request(object_url, tag)
     self
   end
+
+  def object_url
+    "#{service_url}/#{@object_type.underscore.pluralize}/#{@object_id}/tag"
+  end
 end
--- a/app/services/delete_remote_tags.rb
+++ b/app/services/delete_remote_tags.rb
@@ -1,0 +1,10 @@
+class DeleteRemoteTags < RemoteTaggingService
+  def process(tag)
+    post_request(object_url, tag)
+    self
+  end
+
+  def object_url
+    "#{service_url}/#{@object_type.underscore.pluralize}/#{@object_id}/untag"
+  end
+end
--- a/app/services/delete_remote_tags.rb
+++ b/app/services/delete_remote_tags.rb
@@ -1,6 +1,6 @@
 class DeleteRemoteTags < RemoteTaggingService
   def process(tag)
-    post_request(object_url, tag)
+    post_request(object_url, [tag])
     self
   end
 
--- a/app/services/get_remote_tags.rb
+++ b/app/services/get_remote_tags.rb
@@ -7,7 +7,7 @@ class GetRemoteTags < RemoteTaggingService
 
   def process
     response = get_request(object_url)
-    @tags = JSON.parse(response.body)['data'].collect {|tag| {:tag => tag['tag'] } }
+    @tags = JSON.parse(response.body)['data'].collect { |tag| { :tag => tag['tag'] } }
     self
   end
 end
--- a/app/services/get_remote_tags.rb
+++ b/app/services/get_remote_tags.rb
@@ -1,3 +1,4 @@
+require_relative 'mixins/tag_mixin'
 class GetRemoteTags < RemoteTaggingService
   attr_reader :tags
   def initialize(options)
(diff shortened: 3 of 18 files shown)


app/services/workflow_find_service.rb
```diff
@@ -26,8 +26,6 @@ def find_by_tag_resources(tag_resources)
   end
 
   def fq_tag_names(tag_attrs)
-    GetRemoteTags.new(tag_attrs).process.tags.collect do |tag|
-      "/#{tag[:namespace]}/#{tag[:name]}=#{tag[:value]}"
-    end
+    GetRemoteTags.new(tag_attrs).process.tags.collect { |tag| tag[:tag] }
```
hsong-rh: You sure you need an another level of calling `tag` attribute? It's already called in `GetRemoteTags.new(tag_attrs).process.tags`

app/services/get_remote_tags.rb
```diff
@@ -7,14 +7,7 @@ def initialize(options)
 
   def process
     response = get_request(object_url)
-    build_tags(response.body)
+    @tags = JSON.parse(response.body)['data'].collect { |tag| { :tag => tag['tag'] } }
```
bzwei: What is the data type of `JSON.parse(response.body)['data']`? Is it already an array? Do you simply convert the key from `'tag'` to `:tag`?

app/services/workflow_find_service.rb
```diff
@@ -7,13 +7,13 @@ def find(tag_attrs)
...
 
-  # find workflows from a collection of [app_name, object_type, [namespace, key, value]]
+  # find workflows from a collection of [app_name, object_type, [tags]]
   def find_by_tag_resources(tag_resources)
     return [] if tag_resources.blank?
 
     query = nil
     tag_resources.each_with_index do |tr, i|
-      tag_names = tr['tags'].map { |tag| "/#{tag['namespace']}/#{tag['name']}=#{tag['value']}" }
+      tag_names = tr['tags'].collect { |tag| tag[:tag] }
```
bzwei: not sure why we have `tag[:tag]`. shouldn't be `tag['tag']`?
Don't understand why we need to have format of `[{"tag": "tag1"}, {"tag": "tag2"}]`, which can be simply `["tag1", "tag2"]`. I realize this format comes from external services, but still don't understand why.

spec/controllers/v1.0/requests_controller_spec.rb
```diff
@@ -459,7 +459,7 @@
       [{
         'app_name'    => 'app1',
         'object_type' => 'otype1',
-        'tags'        => [{'namespace' => 'ns1', 'name' => 'name1', 'value' => 'v1'}]
+        'tags'        => [{:tag => '/ns1/name1=v1'}]
```
bzwei: prefer `'tag'` other than `:tag`

bzwei: I would like to define request body to have `"tags": ["/ns1/name1=v1", "/ns2/name2=v2"]`
I understand the app that prepares the request body get tags from other API with the format `[{"tag": "/ns1/name1=v1"}, {"tag":"/ns2/name2=v2"}]`. Approval-api has to accept the redundant format if it does not want the caller to have the burden to do a format transformation.

mkanoor: @bzwei @hsong-rh 
The multiple tag is  from the Tagging changes implemented here
https://github.com/RedHatInsights/topological_inventory-api/blob/784bc2640efd190d3c43b3ce5ebaa16f6125906a/public/doc/openapi-3-v2.0.0.json#L10696

app/services/add_remote_tags.rb
```diff
@@ -1,6 +1,10 @@
 class AddRemoteTags < RemoteTaggingService
   def process(tag)
-    post_request(object_url, tag)
+    post_request(object_url, [tag])
```
bzwei: The class name `AddRemoteTags`, and the change here indicates a single tag. Should we change the class name or allow input to be an array?

app/services/delete_remote_tags.rb
```diff
@@ -0,0 +1,10 @@
+class DeleteRemoteTags < RemoteTaggingService
```
bzwei: Same question about the class name vs input argument.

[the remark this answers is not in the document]
mkanoor: @hsong-rh They are different its using inheritance to share the common code. The object_url keeps changing based on what your action is

app/services/remote_tagging_service.rb
```diff
@@ -36,7 +41,7 @@ def self.sources_url
   private_class_method :sources_url
 
   def object_url
-    "#{service_url}/#{@object_type.underscore.pluralize}/#{@object_id}/tags"
+    "#{service_url}/#{@object_type.underscore.pluralize}/#{@object_id}/tags?limit=#{QUERY_LIMIT}&#{approval_tag_filter}"
```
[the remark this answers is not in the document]
mkanoor: @bzwei @hsong-rh The object_url keeps changing based on the action when we want to add_remote_tags the end point ends in "tag" when its being deleted it ends in "untag"

mkanoor:
@bzwei @hsong-rh 
Please review

app/services/add_remote_tags.rb
mkanoor: @bzwei we are adding one tag at a time when we do the link and also when we do the unlink. So its still singular for us might be plural for remote services

app/services/delete_remote_tags.rb
mkanoor: @bzwei We are unlink one workflow at a time, when we unlink multiples we can can change the name

mkanoor: @bzwei Missed it I will fix the inputs

PR #237, opened 2019-12-17
Switch context to get group name

bzwei:
As for now we still need to act as an org_admin in order to get a group name from RBAC service.

### DIFF
--- a/app/services/request_create_service.rb
+++ b/app/services/request_create_service.rb
@@ -42,7 +42,11 @@ class RequestCreateService
   end
 
   def update_leaf_with_workflow(leaf_request, workflow_id, group_ref)
-    group_name = Group.find(group_ref).name if group_ref
+    group_name = if group_ref
+                   ContextService.new(leaf_request.context).as_org_admin do
+                     Group.find(group_ref).name
+                   end
+                 end
 
     leaf_request.update!(:workflow_id => workflow_id, :group_ref => group_ref, :group_name => group_name)
   end


app/services/request_create_service.rb
```diff
@@ -42,7 +42,12 @@ def create_child_requests(request)
   end
 
   def update_leaf_with_workflow(leaf_request, workflow_id, group_ref)
-    group_name = Group.find(group_ref).name if group_ref
+    group_name = nil
+    if group_ref
+      ContextService.new(leaf_request.context).as_org_admin do
+        group_name = Group.find(group_ref).name
+      end
+    end
```
hsong-rh: Here we call RBAC once, just before calling BPM, we will do another. I am thinking is it possible to combine these two calls together?

gmcculloug: Not sure if you like this better, but you can do the same thing without having to default `group_name ` both inside and outside the block:

```ruby
group_name = if group_ref
  ContextService.new(leaf_request.context).as_org_admin do
    Group.find(group_ref).name
  end
end
```

bzwei: In theory we cab have workflows that do NOT call bpm although currently each workflow having a group actually calls bpm.  
Since the second RBAC call runs in background I am not worried about the performance.

gmcculloug: @bzwei rubocop is complaining because it wants it indented like this:
```ruby
group_name = if group_ref
               ContextService.new(leaf_request.context).as_org_admin do
                 Group.find(group_ref).name
               end
             end
```

PR #238, opened 2019-12-18
Enabled bpm service for tag based approval

hsong-rh:
This PR depends on https://github.com/RedHatInsights/approval-process/pull/22

### DIFF
--- a/app/controllers/api/v1/stageaction_controller.rb
+++ b/app/controllers/api/v1/stageaction_controller.rb
@@ -39,9 +39,9 @@ module Api
           comments = params[:message]
         end
 
-        Insights::API::Common::Request.with_request(@stage.request.context.transform_keys(&:to_sym)) do
-          ActsAsTenant.with_tenant(Tenant.find(@stage.tenant_id)) do
-            ActionCreateService.new(@stage.id).create(
+        Insights::API::Common::Request.with_request(@request.context.transform_keys(&:to_sym)) do
+          ActsAsTenant.with_tenant(Tenant.find(@request.tenant_id)) do
+            ActionCreateService.new(@request.id).create(
               :operation    => operation.downcase,
               :processed_by => @approver,
               :comments     => comments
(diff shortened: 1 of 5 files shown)


spec/controllers/v1.0/requests_controller_spec.rb
```diff
@@ -429,7 +429,7 @@
 
   # Test suite for GET /requests/:request_id/requests
   describe 'GET /requests/:request_id/requests' do
-    let(:parent_request) { create(:request, :name => "parent", :owner => "jdoe", :number_of_children => 2, :number_of_finished_children => 0, :tenant_id => tenant.id) }
+    let!(:parent_request) { create(:request, :name => "parent", :owner => "jdoe", :number_of_children => 2, :number_of_finished_children => 0, :tenant_id => tenant.id) }
```
bzwei: why the change is required? there is no other change in this spec

app/controllers/api/v1/stageaction_controller.rb
```diff
@@ -73,15 +73,15 @@ def set_stage
...
 
       def set_order
-        request = @stage.request
+        request = @request
         {
           :orderer       => request.requester_name,
-          :product       => request.content["product"],
-          :portfolio     => request.content["portfolio"],
-          :order_id      => request.content["order_id"],
+          :product       => request.request_context.content["product"],
```
bzwei: you can still use `request.content`

app/controllers/api/v1/stageaction_controller.rb
```diff
@@ -39,9 +39,9 @@ def update
           comments = params[:message]
         end
 
-        Insights::API::Common::Request.with_request(@stage.request.context.transform_keys(&:to_sym)) do
-          ActsAsTenant.with_tenant(Tenant.find(@stage.tenant_id)) do
-            ActionCreateService.new(@stage.id).create(
+        Insights::API::Common::Request.with_request(@request.request_context.context.transform_keys(&:to_sym)) do
```
bzwei: you can access `@request.context`

bzwei:
can we rename `stageaction_controller` to `requestaction_controller`?

spec/controllers/v1.0/requests_controller_spec.rb
hsong-rh: This change fixed the previous spec test failure.

hsong-rh:
I feel it's better to keep it as `stageaction_controller`, in sync with the endpoint `/api/v1.0/stageaction`, which is specialized for us.

PR #239, opened 2019-12-19
Added more description details on /requests/:id/requests endpoint

hsong-rh:


public/doc/openapi-3-v1.0.0.json
```diff
@@ -331,7 +331,7 @@
                   "Request"
                 ],
                 "summary": "Return an array of child requests of a given request id",
-                "description": "Return an array of child requests of a given request id, available for admin/requester",
+                "description": "Return an array of child requests of a given request id. The result depends on the x-rh-persona header (approval/admin, approval/requseter, or approval/approver).",
                 "operationId": "listRequestsByRequest",
                 "parameters": [
```
bzwei: We need to copy https://github.com/RedHatInsights/approval-api/blob/master/public/doc/openapi-3-v1.0.0.json#L207

bzwei: For `"responses"` part we need to copy https://github.com/RedHatInsights/approval-api/blob/master/public/doc/openapi-3-v1.0.0.json#L224

PR #240, opened 2019-12-20
Fix the input parameter to match the class names

mkanoor:
In an earlier PR I missed the fact that the class name
was plural but the input parameters were singular. Fixed
input parameters so that its consistent with class name

Missed from PR https://github.com/RedHatInsights/approval-api/pull/235

### DIFF
--- a/app/services/add_remote_tags.rb
+++ b/app/services/add_remote_tags.rb
@@ -1,6 +1,6 @@
 class AddRemoteTags < RemoteTaggingService
-  def process(tag)
-    post_request(object_url, [tag])
+  def process(tags)
+    post_request(object_url, tags)
     self
   end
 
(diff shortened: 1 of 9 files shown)


app/services/add_remote_tags.rb
```diff
@@ -1,6 +1,6 @@
 class AddRemoteTags < RemoteTaggingService
-  def process(tag)
-    post_request(object_url, [tag])
+  def process(tags)
+    post_request(object_url, tags)
```
gmcculloug: @mkanoor You could also do this as `Array(tags)` which would leave an array alone and wrap a non-array element in an array.  Then the caller does not have to always pass an array for a single element.  Same with `DeleteRemoteTags`.

mkanoor: @gmcculloug Then it doesn't match the class name, we would have to call the class to indicate that it can take a scalar or an array argument

gmcculloug: Not sure I fully agree about having to rename the class and is weird to force the caller to artificially create an array to pass to the method.  I'll try to find some good examples, but we can deal with that at a later time.  

I would rather move forward with the fix and discuss later.

PR #242, opened 2019-12-20
Suppressed log messages

hsong-rh:
Reduced log messages to improve performance.

### DIFF
--- a/app/controllers/api/v1/actions_controller.rb
+++ b/app/controllers/api/v1/actions_controller.rb
@@ -43,7 +43,7 @@ module Api
 
         # Only approver can reach here
         action_ids = approver_id_list(relation.model.table_name)
-        Rails.logger.info("approver scope for actions: #{action_ids}")
+        Rails.logger.debug { "Approver scope for actions: #{action_ids}" }
 
         relation.where(:id => action_ids)
       end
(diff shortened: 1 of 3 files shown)


app/controllers/api/v1/actions_controller.rb
```diff
@@ -43,7 +43,7 @@ def rbac_scope(relation)
 
         # Only approver can reach here
         action_ids = approver_id_list(relation.model.table_name)
-        Rails.logger.info("approver scope for actions: #{action_ids}")
+        Rails.logger.info { "Number of approver scope for actions: #{action_ids.count}" }
```
bzwei: either `debug { }` or `info()`
If the actual IDs are needed for debugging, you can still log all `action_ids` with `debug {}`.

app/controllers/api/v1/requests_controller.rb
```diff
@@ -51,7 +51,7 @@ def rbac_scope(relation)
         # for admin
         return relation unless ids
 
-        Rails.logger.info("Accessible #{relation.model.table_name} ids: #{ids}")
+        Rails.logger.debug { "Accessible #{relation.model.table_name} ids: #{ids}" }
```
gmcculloug: This is the log line I saw posting 9400+ IDs.  That will use up some time to write out.  Hoping this helps.

hsong-rh: Then I will remove this message, just in the future real case has the same scenario.

PR #243, opened 2020-01-02
Create access_control_entries table to store ACLs locally

bzwei:
For the purpose of minimizing the number of remote calls to RBAC service, we brought in similar changes from https://github.com/RedHatInsights/catalog-api/pull/500. Instance level ACLs are now stored in local database. 
Class level permissions for approver role are also locally managed.


app/services/workflow_update_service.rb
```diff
@@ -6,27 +6,10 @@ def initialize(workflow_id)
...
-        end
-      end
-
-      if removed_group_refs.any?
-        ContextService.new(Insights::API::Common::Request.current.to_h.transform_keys(&:to_s)).as_org_admin do
-          aps.remove_resource_from_groups(workflow.id, removed_group_refs)
-        end
+    options[:access_control_entries] =
+      (options[:group_refs] || []).collect do |uuid|
+        AccessControlEntry.new(:group_uuid => uuid, :permission => 'approve')
```
hsong-rh: We should remove the old `AccessControlEntry` here, right? Otherwise the following lookup based on local acls will return out of sync records.

bzwei: The `update` method automatically removes outdated associations.

hsong-rh: 👍

app/controllers/api/v1/mixins/rbac_mixin.rb
```diff
@@ -121,38 +125,31 @@ def owner_request_ids
...
         end
 
         # The access list regular requesters have
         def requester_acls
           Insights::API::Common::RBAC::ACL.new.create(nil, OWNER_PERMISSIONS)
         end
+
+        # Access list approvers can have
+        def approver_acls
+          Insights::API::Common::RBAC::ACL.new.create(nil, APPROVER_PERMISSIONS)
```
mkanoor: @bzwei Do we need to be using this class, in catalog we are getting rid of this class from the common gem.
Since the ACE's are stored locally. You only need permissions from the RBAC service now

mkanoor: https://github.com/RedHatInsights/insights-api-common-rails/blob/master/lib/insights/api/common/rbac/access.rb

mkanoor: We are trying to get rid of this one
https://github.com/RedHatInsights/insights-api-common-rails/blob/master/lib/insights/api/common/rbac/acl.rb

Which did all of the heavy lifting for managing ACL's on the RBAC server side

mkanoor: We use this
https://github.com/RedHatInsights/catalog-api/blob/cebf916601c3df35e15c5fbb53a748458e86578e/app/controllers/api/v1/mixins/rbac_mixin.rb#L37
Which does the resolution on the RBAC service and gets us the base Access Object.

mkanoor: Or
https://github.com/RedHatInsights/catalog-api/blob/cebf916601c3df35e15c5fbb53a748458e86578e/app/controllers/api/v1/mixins/index_mixin.rb#L33

PR #245, opened 2020-01-06
Disable filtering when make GET requests to get tags.

mkanoor:
Since tagging was re-implemented it has ony one attibute called
:tag, so its difficult to filter based on internal variables for
Tags(:namespace, :name, :value)

### DIFF
--- a/app/services/mixins/tag_mixin.rb
+++ b/app/services/mixins/tag_mixin.rb
@@ -9,8 +9,4 @@ module TagMixin
   def approval_tag(workflow_id)
     { 'tag' => fq_tag_name(workflow_id) }
   end
-
-  def approval_tag_filter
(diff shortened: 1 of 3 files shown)


app/services/mixins/tag_mixin.rb
```diff
@@ -9,8 +9,4 @@ def fq_tag_name(workflow_id)
   def approval_tag(workflow_id)
     { 'tag' => fq_tag_name(workflow_id) }
   end
-
-  def approval_tag_filter
```
mkanoor: This was dead code, the escape never worked so I had to pass them as separate params and have Faraday do the encoding of the params.

mkanoor:
https://github.com/RedHatInsights/catalog-api/pull/533

PR #251, opened 2020-01-08
Fix Workflow#external_processing? and #external_signal?

bzwei:
This is a bug. Both method failed for `Always Approve` workflow because it does not have a template.

### DIFF
--- a/spec/models/workflow_spec.rb
+++ b/spec/models/workflow_spec.rb
@@ -98,4 +98,23 @@ RSpec.describe Workflow, :type => :model do
       end
     end
   end
+
+  describe '#external_signal? & #external_processing?' do
+    context 'no template' do
+      it 'has no external_processing' do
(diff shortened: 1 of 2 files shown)


spec/models/workflow_spec.rb
```diff
@@ -98,4 +98,23 @@
       end
     end
   end
+
+  describe '#external_signal? & #external_processing?' do
+    context 'no template' do
+      it 'has no external_processing' do
+        expect(workflow.external_processing?).to be_falsy
```
mkanoor: @bzwei Should this be falsey or falsy?

mkanoor: https://relishapp.com/rspec/rspec-expectations/docs/built-in-matchers/be-matchers#be-falsey-matcher

bzwei: should be `falsey`. Why `falsy` also works?

PR #260, opened 2020-01-21
Update CMD script to only start the rails server if the migrations succeed

lindgrenj6:
https://projects.engineering.redhat.com/browse/SSP-1150

Following the same thing I did on Catalog, update the CMD script so the container will only start if the datase migrations (and seeding in this case) succeed. 

Related Catalog PR: https://github.com/RedHatInsights/catalog-api/pull/563

### DIFF
--- a/docker-assets/run_rails_server
+++ b/docker-assets/run_rails_server
@@ -10,5 +10,4 @@ KEY
 
 write_encryption_key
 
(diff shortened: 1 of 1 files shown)


lindgrenj6:
cc @gmcculloug @syncrou @carbonin

PR #265, opened 2020-01-29
More restrict action operation validation

bzwei:
Refactor the operation validation based on user's role.
Allow `notify` operation with header `x-rh-random-access-key`. This operation is not documented for public use in openapi spec. It is designed to be used as an internal call from PAM.


bzwei:
depend on https://github.com/RedHatInsights/approval-process/pull/29

PR #270, opened 2020-02-05
Update Dockerfile to use UBI instead of CentOS

lindgrenj6:
https://projects.engineering.redhat.com/browse/SSP-1149

Update Dockerfile to use a RHEL based image rather than `manageiq/ruby` which was based on CentOS. This is for our security audit in the future. 

Tested this in my own namespace with my own database and had great success:
image

cc @gmcculloug @carbonin @gtanzillo exhibit A of converting our services to use UBI and thus RHEL based images.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,27 +1,25 @@
-FROM manageiq/ruby:latest
+FROM registry.access.redhat.com/ubi8/ubi
 
-RUN yum -y install centos-release-scl-rh && \
-    yum -y install --setopt=tsflags=nodocs \
-                   # To compile native gem extensions
-                   gcc-c++ \
-                   git \
-                   # To compile pg gem
-                   rh-postgresql10-postgresql-devel \
-                   rh-postgresql10-postgresql-libs \
-                   && \
+RUN dnf -y --disableplugin=subscription-manager module enable ruby:2.5 && \
+    dnf -y --disableplugin=subscription-manager --setopt=tsflags=nodocs install \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,26 +1,16 @@
...
-    yum clean all
+FROM registry.access.redhat.com/ubi8/ruby-25
+# This image runs as non-root by default.
+USER 0
 
 ENV WORKDIR /opt/insights-api-service_approval/
 WORKDIR $WORKDIR
 
 COPY Gemfile $WORKDIR
-RUN source /opt/rh/rh-postgresql10/enable && \
```
carbonin: :heart: Not a fan of the scl generally

Dockerfile
```diff
@@ -1,26 +1,16 @@
...
-                   gcc-c++ \
-                   git \
-                   # To compile pg gem
-                   rh-postgresql10-postgresql-devel \
-                   rh-postgresql10-postgresql-libs \
-                   && \
-    yum clean all
+FROM registry.access.redhat.com/ubi8/ruby-25
+# This image runs as non-root by default.
+USER 0
```
carbonin: If it doesn't run as root, what is the "right" way to build using it? Surely every dockerfile can't be expected to include this line, right?

Or is this some kind of s2i thing?

lindgrenj6: Yep it's all about the fact that this is also a s2i builder image. Switching this over to using the regular `ubi8/ubi` image fixed the problem.

PR #271, opened 2020-02-05
Add error operation to action, and failed state to request

bzwei:
When an error or exception is found at approval processing runtime, an error action can be added and flag the request as failed state.

It skips any un-notified sub requests. A request completed event will be sent to Kafka with decision to be `error` and `reason` including the detailed error message.


app/services/action_create_service.rb
```diff
@@ -83,13 +83,18 @@ def deny(comments)
...
+    raise Exceptions::InvalidStateTransitionError, "The request has already finished" if request.finished?
 
     {:state => Request::CANCELED_STATE, :decision => Request::CANCELED_STATUS}.tap do |h|
       h[:reason] = comments if comments
     end
   end
+
+  def error(comments)
+    raise Exceptions::InvalidStateTransitionError, "Current request has already finished" if request.finished?
+    raise Exceptions::ApprovalError, "Failure reason is missing" unless comments
```
mkanoor: @bzwei Should this be listed here
https://github.com/RedHatInsights/approval-api/blob/c5e99a5c5ad38597748309e872a7190df8247971/config/initializers/custom_exception_mappings.rb#L8
Also are Approval Errors bad request or server errors. A bad request would mean that the user entered something bad in the dialogs but a server internal error would mean the users request is OK but something is broken in the back end. A server error 5xx can trigger some cleanup activity on the part of the Administrator, who might have to fix his Approval workflow e.g. remove some stale groups.

app/services/action_create_service.rb
```diff
@@ -75,21 +75,26 @@ def deny(comments)
     unless request.state == Request::NOTIFIED_STATE
       raise Exceptions::InvalidStateTransitionError, "Current request is not in notified state"
     end
-    raise Exceptions::ApprovalError, "Reason to deny the request is missing" unless comments
+    raise Exceptions::UserError, "Reason to deny the request is missing" unless comments
```
mkanoor: @bzwei This error message is a bit confusing, is it saying that the Approver didn't give a reason why he denied the request.
Can the message be "A reason has to be provided if a request is being denied" or
"The approver has not specified a reason about why a request has been denied"

PR #275, opened 2020-02-07
Add /health for liveness probe

lindgrenj6:
Following the pattern created in the common gem, catalog PR for reference: https://github.com/RedHatInsights/catalog-api/pull/528/files

This will allow us to add a livenessProbe to the deployment config so the pod will auto restart if something is wrong.


lindgrenj6:
cc @gmcculloug

PR #278, opened 2020-02-12
Added version 1x1 for openapi json error state

syncrou:
Exposes v1x1 versioning for the error state introduced in #271


syncrou:
@bzwei @hsong-rh this is ready to be reviewed now.

public/doc/openapi-3-v1.1.json
```diff
@@ -3,7 +3,7 @@
     "info": {
         "title": "Insights Service Approval APIs",
         "description": "APIs to query approval service",
-        "version": "1.0.0",
+        "version": "1.1.0",
```
bzwei: should this be `1.1.0` or `1.1`?

spec/support/v1_helper.rb
```diff
@@ -0,0 +1,5 @@
+module V1Helper
```
bzwei: why the file name is asymmetric? should it be `v1x0_helper.rb`?

spec/spec_helper.rb
```diff
@@ -2,21 +2,6 @@
 Dir[Insights::API::Common::Engine.root.join("spec/support/**/*.rb")].each { |f| require f }
 require 'webmock/rspec'
 
-# This file was generated by the `rails generate rspec:install` command. Conventionally, all
```
bzwei: Is this file auto generated or hand edited? From the deleted comments it seemed to be auto generated.

spec/controllers/v1/root_controller_spec.rb
```diff
@@ -1,8 +1,7 @@
-RSpec.describe Api::V1x0::RootController, :type => :request do
+RSpec.describe Api::V1x0::RootController, :type => [:request, :v1x1] do
```
bzwei: Why here `Api::V1x0` does not match `[:request, :v1x1]`?

bzwei:
@syncrou can you explain the relation between v1x0, v1x1 and v1? It seems to me v1 is the same as v1x0. Why?

spec/controllers/v1/root_controller_spec.rb
syncrou: I moved that root controller to a major version only path.  Because that spec always should return the highest minor version.

spec/spec_helper.rb
syncrou: The original is auto generated. I just pulled out all the commented out code to clean up the file.

syncrou:
You are correct. v1x0 is the same as v1. Since in the app/controllers we already had v1 and v1x0, I left those entries in the specs.  I do agree we could probably clean that up, but that would have made this PR very difficult to read. 

In reality though, using v1x0 vs v1 is just symantics.  The two really are the exact same. The main part is that if a url redirects correctly: /api/v1/endpoint -> /api/<latest minor version>/endpoint

So we cloud rewrite all the namespaces for v1 to be v1x0 for consistency but code wise it changes nothing.

spec/support/v1_helper.rb
syncrou: https://github.com/RedHatInsights/approval-api/pull/278#issuecomment-593497675

PR #282, opened 2020-02-13
precompile assets

bzwei:


bzwei:
@lindgrenj6 Should we do this in the Dockerfile so that the assets does not need to be generated every time when the pod is booted?

docker-assets/run_rails_server
```diff
@@ -10,4 +10,4 @@ KEY
 
 write_encryption_key
 
-bundle exec rake db:migrate db:seed && bundle exec rails server
+bundle exec rake db:migrate db:seed && bundle exec rails assets:precompile && bundle exec rails server
```
lindgrenj6: instead of adding a third bundle exec here, I would suggest changing it to this:
`bundle exec rake db:migrate db:seed assets:precompile && bundle exec rails server`

It takes a long time to spin up for `bundle exec`, so doing it only twice is still the best solution.

PR #284, opened 2020-02-14
Change asset path so fonts/css load correctly

lindgrenj6:
Laura reached out to me why the CSS wasn't loading from the Rails application, tried just turning on the `public_file_server` but that wasn't enough, so I changed the assets prefix as well. 

This way it generates the links in the UI for the fonts/css correctly.

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -6,6 +6,10 @@ Rails.application.configure do
   # since you don't have to restart the web server when you make code changes.
   config.cache_classes = false
 
+  config.public_file_server.enabled = true
+  # sets the asset path to either `/api/static` or `/api/approval/static`
+  config.assets.prefix = File.join(File::SEPARATOR, 'api', ENV['APP_NAME'].to_s, 'static')
+
   # Do not eager load code on boot.
   config.eager_load = false
 
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -18,9 +18,9 @@ Rails.application.configure do
   config.consider_all_requests_local       = false
   config.action_controller.perform_caching = true
 
-  # Disable serving static files from the `/public` folder by default since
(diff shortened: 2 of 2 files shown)


lindgrenj6:
cc @lgalis

config/environments/development.rb
```diff
@@ -6,6 +6,10 @@
   # since you don't have to restart the web server when you make code changes.
   config.cache_classes = false
 
+  config.public_file_server.enabled = true
+  # sets the asset path to either `/api/static` or `/api/approval/static`
+  config.assets.prefix = "/api#{ENV['APP_NAME'] ? "/#{ENV['APP_NAME']}/" : "/"}static"
```
gmcculloug: @lindgrenj6 Would suggest using `File.join`:
```ruby
config.assets.prefix = File.join(File::SEPARATOR, 'api', ENV['APP_NAME'].to_s, 'static')
```
if `ENV['APP_NAME']` is `nil` it will not be included in the path.

config/environments/production.rb
```diff
@@ -18,9 +18,9 @@
   config.consider_all_requests_local       = false
   config.action_controller.perform_caching = true
 
-  # Disable serving static files from the `/public` folder by default since
-  # Apache or NGINX already handles this.
-  config.public_file_server.enabled = ENV['RAILS_SERVE_STATIC_FILES'].present?
+  config.public_file_server.enabled = true
+  # sets the asset path to either `/api/static` or `/api/approval/static`
+  config.assets.prefix = "/api#{ENV['APP_NAME'] ? "/#{ENV['APP_NAME']}/" : "/"}static"
```
gmcculloug: same here.

config/environments/development.rb
lindgrenj6: Updated to use this instead, works great!

PR #288, opened 2020-02-26
Remove Always approve workflow

bzwei:
https://projects.engineering.redhat.com/browse/SSP-856. 
If no workflow is linked to any resource in an order, the default is already always approve. There is no need to ask a user to specifically choose an always approve workflow.

### DIFF
--- a/db/migrate/20200226115345_delete_always_approve_from_workflows.rb
+++ b/db/migrate/20200226115345_delete_always_approve_from_workflows.rb
@@ -1,0 +1,13 @@
+class DeleteAlwaysApproveFromWorkflows < ActiveRecord::Migration[5.2]
+  def up
+    always_approve = Workflow.find_by(:name => 'Always approve')
+    return unless always_approve
+
+    Request.where(:workflow => always_approve).update_all(:workflow_id => nil)
+    always_approve.delete
+  end
+
+  def down
+    Workflow.create(:name => 'Always approve', :description => 'Always auto approve by system. No approvers are assigned.')
+  end
+end
--- a/spec/controllers/v1.0/graphql_controller_spec.rb
+++ b/spec/controllers/v1.0/graphql_controller_spec.rb
@@ -6,7 +6,7 @@ RSpec.describe Api::V1x0::GraphqlController, :type => :request do
 
   let!(:template) { create(:template) }
   let(:template_id) { template.id }
-  let!(:workflows) { create_list(:workflow, 4, :template_id => template.id) }
+  let!(:workflows) { create_list(:workflow, 4, :template => template, :tenant => tenant) }
   let(:id) { workflows.first.id }
   let(:roles_obj) { instance_double(Insights::API::Common::RBAC::Roles, :roles => roles) }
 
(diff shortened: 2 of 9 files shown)


db/migrate/20200226115345_delete_always_approve_from_workflows.rb
```diff
@@ -0,0 +1,13 @@
+class DeleteAlwaysApproveFromWorkflows < ActiveRecord::Migration[5.2]
+  def up
+    always_approve = Workflow.find_by(:name => 'Always approve')
+    return unless always_approve
+
+    Request.where(:workflow_id => always_approve.id).update_all(:workflow_id => nil)
```
gmcculloug: You can let Rails do some of the work for you here but referencing the object and not the `id`.

```ruby
Request.where(:workflow => always_approve).update_all(:workflow => nil)
```

spec/controllers/v1.0/graphql_controller_spec.rb
```diff
@@ -6,7 +6,7 @@
 
   let!(:template) { create(:template) }
   let(:template_id) { template.id }
-  let!(:workflows) { create_list(:workflow, 4, :template_id => template.id) }
+  let!(:workflows) { create_list(:workflow, 4, :template_id => template.id, :tenant_id => tenant.id) }
```
gmcculloug: Same with these specs, you just need to reference the objects:
```ruby
let!(:workflows) { create_list(:workflow, 4, :template => template, :tenant => tenant) }
```

spec/controllers/v1.0/requests_controller_spec.rb
```diff
@@ -8,7 +8,7 @@
   let(:headers_with_approver)  { default_headers.merge(Insights::API::Common::Request::PERSONA_KEY => described_class::PERSONA_APPROVER) }
   let(:headers_with_requester) { default_headers.merge(Insights::API::Common::Request::PERSONA_KEY => described_class::PERSONA_REQUESTER) }
 
-  let(:workflow) { create(:workflow, :name => 'Test always approve') }
+  let(:workflow) { create(:workflow, :name => 'Test always approve', :tenant_id => tenant.id) }
```
gmcculloug: same

spec/controllers/v1.0/workflows_controller_spec.rb
```diff
@@ -5,7 +5,7 @@
 
   let!(:template) { create(:template) }
   let(:template_id) { template.id }
-  let!(:workflows) { create_list(:workflow, 16, :template_id => template.id) }
+  let!(:workflows) { create_list(:workflow, 16, :template_id => template.id, :tenant_id => tenant.id) }
```
gmcculloug: same

db/migrate/20200226115345_delete_always_approve_from_workflows.rb
bzwei: Seems `update_all` does not follow relationships

gmcculloug: Ok, but we can still update the `where` statement.

PR #289, opened 2020-02-27
Changed to handle tenant from request

hsong-rh:
Changed to use tenant from request, similar work as following PR:
https://github.com/RedHatInsights/catalog-api/pull/603


hsong-rh:
@mkanoor @bzwei Please review.

PR #292, opened 2020-03-02
Decrypt url to get random access key and approver name

bzwei:
https://projects.engineering.redhat.com/browse/SSP-1290

The key and name are encrypted by https://github.com/RedHatInsights/approval-process/pull/34
This is the part to decrypt them.


bzwei:
Will take another approach.

PR #297, opened 2020-03-04
Move random_access_key into a separate table

bzwei:
Each approver per request has a unique random_access_key

https://projects.engineering.redhat.com/browse/SSP-1313

Before the url looks like https://ci.cloud.redhat.com/api/approval/v1.0/stageaction/ec0150c649d9b5fbfb74439eeaaffc35?approver=Z21jY3VsbG9AcmVkaGF0LmNvbQ==

The new url looks like
https://ci.cloud.redhat.com/api/approval/v1.0/stageaction/ec0150c649d9b5fbfb74439eeaaffc35

### DIFF
--- a/app/models/random_access_key.rb
+++ b/app/models/random_access_key.rb
@@ -1,0 +1,7 @@
+class RandomAccessKey < ApplicationRecord
+  belongs_to :request
+
+  after_initialize do |key|
+    key.access_key ||= SecureRandom.hex(16)
+  end
+end
--- a/app/services/jbpm_process_service.rb
+++ b/app/services/jbpm_process_service.rb
@@ -33,10 +33,25 @@ class JbpmProcessService
       options = {
         'request'         => request,
         'request_context' => request.request_context.as_json,
-        'groups'          => [group].as_json
+        'groups'          => enhance_groups([group].as_json)
       }
     end
     options
+  end
+
+  def enhance_groups(groups_json)
+    random_access_keys = []
+    groups_json.each do |group_json|
+      group_json['users']&.each do |user_json|
+        full_name = "#{user_json['first_name']} #{user_json['last_name']}"
+        random_access_key = RandomAccessKey.new(:approver_name => full_name)
+        user_json['random_access_key'] = random_access_key.access_key
+        random_access_keys << random_access_key
+      end
+    end
+    request.random_access_keys = random_access_keys
+
+    groups_json
   end
 
   def signal_options(decision)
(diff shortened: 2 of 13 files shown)


app/services/jbpm_process_service.rb
```diff
@@ -33,12 +33,27 @@ def process_options
...
       }
     end
     options
   end
 
+  def enhance_groups(groups_json)
+    random_access_keys = []
+    groups_json.each do |group_json|
+      group_json['users']&.each do |user_json|
+        access_key = SecureRandom.hex(16)
```
gmcculloug: If we ever want to extend this value to a longer value all previously created access keys would still continue to work, correct?

app/services/jbpm_process_service.rb
```diff
@@ -33,12 +33,27 @@ def process_options
...
+        user_json['random_access_key'] = access_key
+        full_name = "#{user_json['first_name']} #{user_json['last_name']}"
+        random_access_keys << RandomAccessKey.new(:access_key => access_key, :approver_name => full_name)
```
gmcculloug: I want to suggest a slight change to this logic.  It seems odd that the `RandomAccessKey` class does not create the key, but gets it passed to it.  That class should encapsulate the creation of the key.

Then this method changes to create the `RandomAccessKey` instance first by only passing the `approver_name` and you set `ser_json['random_access_key']` based on the `access_key` value from that instance.

```ruby
def enhance_groups(groups_json)
  groups_json.each do |group_json|
    group_json['users']&.each do |user_json|
      request.random_access_keys << random_access_key = RandomAccessKey.new(:approver_name => full_name)

      user_json['random_access_key'] = random_access_keys.access_key
      full_name = "#{user_json['first_name']} #{user_json['last_name']}"
    end
  end

  groups_json
end
```

app/models/random_access_key.rb
```diff
@@ -0,0 +1,3 @@
+class RandomAccessKey < ApplicationRecord
+  belongs_to :request
```
gmcculloug: Based on my suggest you would create the `access_key` when this method is initialized.  Suggest using a before callback.

app/services/jbpm_process_service.rb
bzwei: yes

app/services/jbpm_process_service.rb
bzwei: Moved the creation of key inside the model class.
I prefer to set `request.random_access_keys` outside the `do` block to optimize sql statements.

bzwei:
Do not merge until the work on PAM side is done.

app/models/random_access_key.rb
```diff
@@ -0,0 +1,7 @@
...
+
+  after_initialize do |key|
```
gmcculloug: @syncrou @eclarizio Is there a better callback to use here?  I'm concerned we are doing this ever time the object is instantiated (which is protected be the `||=`) but it would be nice to avoid that.  I have used `default_value_for` in the past, but that might be overkill for this one property.

gmcculloug:
@bzwei If you rebase you will pick up https://github.com/RedHatInsights/approval-api/commit/235180a8 which resolves the Hakiri failure.

PR #298, opened 2020-03-06
Update request state when KIE server raises errors

hsong-rh:
Update request's state to `failed` when KIE has errors.

### DIFF
--- a/app/services/request_update_service.rb
+++ b/app/services/request_update_service.rb
@@ -18,11 +18,11 @@ class RequestUpdateService
   private
 
   def started(options)
-    request.update!(options)
-
     if request.leaf?
       start_request
     end
+
+    request.update!(options)
 
     EventService.new(request).request_started if request.root?
 
@@ -150,6 +150,7 @@ class RequestUpdateService
 
   # complete the external approval process if configured
   def finish_request(decision)
+    return if request_completed?(decision)
     return unless request.process_ref.present? && request.workflow.try(:external_processing?)
 
     template = request.workflow.template
--- a/app/services/request_update_service.rb
+++ b/app/services/request_update_service.rb
@@ -18,11 +18,11 @@ class RequestUpdateService
   private
 
   def started(options)
+    request.update!(options)
+
     if request.leaf?
       start_request
     end
-
-    request.update!(options)
 
     EventService.new(request).request_started if request.root?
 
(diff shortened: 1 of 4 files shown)


app/services/request_update_service.rb
```diff
@@ -19,13 +19,13 @@ def update(options)
   private
 
   def started(options)
+    request.update!(options)
```
bzwei: does not need to move up.

spec/services/request_create_service_spec.rb
```diff
@@ -51,7 +51,6 @@
             :state             => Request::STARTED_STATE,
             :decision          => Request::UNDECIDED_STATUS,
             :reason            => nil,
-            :random_access_key => be_kind_of(String),
```
bzwei: not from this PR

spec/services/request_create_service_spec.rb
```diff
@@ -76,8 +75,7 @@
           :requester_name    => 'John Doe',
           :owner             => 'jdoe',
           :state             => Request::NOTIFIED_STATE,
-          :decision          => Request::UNDECIDED_STATUS,
-          :random_access_key => be_kind_of(String)
```
bzwei: not from this PR

bzwei:
@hsong-rh need to add spec test when error is raised

spec/services/request_create_service_spec.rb
hsong-rh: Changed for spec test failure.

spec/services/request_create_service_spec.rb
hsong-rh: Same reason as above

app/services/request_update_service.rb
hsong-rh: Has to change the order,  otherwise 'notify' post action will fail

hsong-rh:
@bzwei spec test is added.

app/services/request_update_service.rb
bzwei: We discussed it before. There is no notify action calling into approval-api from approval-pam during `start_request`.

spec/services/request_create_service_spec.rb
bzwei: you should not test with code merged from pending #297

app/services/request_update_service.rb
```diff
@@ -149,7 +149,9 @@ def notify_request
 
   # complete the external approval process if configured
   def finish_request(decision)
-    return unless request.workflow.try(:external_processing?)
+    # PAM sends error and stops the process already
+    return if decision == Action::ERROR_OPERATION
```
bzwei: probably need a better indicator. If the error action was caused by other code and the PAM process is still running we still need to send the signal.
Why `request.process_ref.present?` is not sufficient? If PAM start failed with exception, we should not be able to record its `process_ref`.

hsong-rh: As discussed, we still can have `process_ref` since we turned on `Async` flag. Changed to use `request_complete?` to skip sending signal.

PR #299, opened 2020-03-09
Check permission for both approvers and requesters

bzwei:
https://projects.engineering.redhat.com/browse/SSP-1323

We need to check permission for both approvers and requesters. The permission is granted if either check passes.

### DIFF
--- a/app/controllers/api/v1/mixins/rbac_mixin.rb
+++ b/app/controllers/api/v1/mixins/rbac_mixin.rb
@@ -52,7 +52,7 @@ module Api
         def resource_accessible?(resource, verb)
           return true if admin?
 
-          approver? ? permitted?(APPROVER_PERMISSIONS, resource, verb) : permitted?(OWNER_PERMISSIONS, resource, verb)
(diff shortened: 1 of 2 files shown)


app/controllers/api/v1/mixins/rbac_mixin.rb
```diff
@@ -52,7 +52,7 @@ def permission_check(verb, klass = controller_name.classify.constantize)
         def resource_accessible?(resource, verb)
           return true if admin?
 
-          approver? ? permitted?(APPROVER_PERMISSIONS, resource, verb) : permitted?(OWNER_PERMISSIONS, resource, verb)
+          approver? && permitted?(APPROVER_PERMISSIONS, resource, verb) || permitted?(OWNER_PERMISSIONS, resource, verb)
```
gmcculloug: @bzwei Would it be more performant to swap the order of these checks and do `permitted?(OWNER_PERMISSIONS, resource, verb)` first and `approver? && permitted?(APPROVER_PERMISSIONS, resource, verb)` only if needed?

PR #301, opened 2020-03-11
Use service to service call to RBAC service where required

bzwei:
Refactor `ContentService#act_as_org_admin` method wherever we need to call RBAC with org_admin right to use the newly supported service_to_service call feature from RBAC team.

Wait for a new `insights-api-common` gem release that include https://github.com/RedHatInsights/insights-api-common-rails/pull/168.

### DIFF
--- a/app/services/context_service.rb
+++ b/app/services/context_service.rb
@@ -6,11 +6,19 @@ class ContextService
   end
 
   def with_context(&block)
-    switch_context(context, &block)
+    switch_context(context, false, &block)
   end
 
   def as_org_admin(&block)
-    switch_context(org_admin_context, &block)
+    if ENV['RBAC_PSK']
+      begin
+        switch_context(context, true, &block)
+      ensure
+        Thread.current[:rbac_extra_headers] = nil
+      end
+    else
+      switch_context(org_admin_context, false, &block)
+    end
   end
 
   private
@@ -23,9 +31,21 @@ class ContextService
     end
   end
 
-  def switch_context(new_context)
+  def service_to_service_headers
+    {
+      'x-rh-rbac-psk'       => ENV['RBAC_PSK'],
+      'x-rh-rbac-account'   => ActsAsTenant.current_tenant.external_tenant,
+      'x-rh-rbac-client-id' => 'approval',
+      'x-rh-identity'       => nil
+    }
+  end
+
+  def switch_context(new_context, extra_headers)
     Insights::API::Common::Request.with_request(new_context.transform_keys(&:to_sym)) do |current|
-      ActsAsTenant.with_tenant(Tenant.find_by(:external_tenant => current.tenant)) { yield }
+      ActsAsTenant.with_tenant(Tenant.find_by(:external_tenant => current.tenant)) do
+        Thread.current[:rbac_extra_headers] = service_to_service_headers if extra_headers
+        yield
+     end
     end
   end
 end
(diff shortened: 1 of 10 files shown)


app/controllers/api/v1/mixins/rbac_mixin.rb
```diff
@@ -134,7 +134,7 @@ def visible_request_ids_for_approver
         end
 
         def assigned_group_refs
-          Insights::API::Common::RBAC::Service.call(RBACApiClient::GroupApi) do |api|
+          Insights::API::Common::RBAC::Service.call(RBACApiClient::GroupApi, Thread.current[:rbac_extra_headers] || {}) do |api|
```
mkanoor: @bzwei Shouldn't this happen only on the call where we need to fetch the emails from the groups?
Instead of using the Thread.current which can get reused for different transactions, can we be explicit on which calls are admin calls. Maybe even name them as admin calls, so its easy to track them. Plus Admin calls need those extra headers, if its not there its an error.

bzwei: My intention is to refactor `ContentService#act_as_org_admin`. All RBAC calls by default do not use service_to_service headers. They are set only through `act_as_org_admin` and get cleared immediately after.

app/services/context_service.rb
```diff
@@ -23,6 +32,19 @@ def org_admin_context
...
+  def service_to_service_headers
+    {
+      'x-rh-rbac-psk'       => ENV['RBAC_PSK'],
+      'x-rh-rbac-account'   => account_number,
+      'x-rh-rbac-client-id' => 'approval'
+    }
+  end
+
+  def account_number
+    Insights::API::Common::Request.current&.tenant ||
```
mkanoor: @bzwei Is this for testing? Because you would always have the Insights::API::Common::Request.current.tenant because we are storing the context with every request in the database so when we have to restart something we would first initialize the request and then call the controller code.

bzwei: yes. it is very useful if we need to debug through rails console

app/services/context_service.rb
```diff
@@ -23,6 +32,20 @@ def org_admin_context
     end
   end
 
+  def service_to_service_headers
+    {
+      'x-rh-rbac-psk'       => ENV['RBAC_PSK'],
+      'x-rh-rbac-account'   => account_number,
+      'x-rh-rbac-client-id' => 'approval',
+      'x-rh-identity'       => nil
```
mkanoor: @bzwei wouldn't this conflict with 
https://github.com/RedHatInsights/insights-api-common-rails/blob/064511ec143d299c0a3a271dfd0ea24e6f86965d/lib/insights/api/common/rbac/service.rb#L59 we will set the x_rh_identity to nil, not sure if that is not going to get sent over, we might need a compact in this code to remove the nil entries.

bzwei: I tested and it works

app/services/context_service.rb
mkanoor: @bzwei For Testing acts_as_tenant has a simple way to do this from their console
```
# This manually sets the current tenant for testing purposes. In your app this is handled by the gem.
ActsAsTenant.current_tenant = Account.find(3)
```

bzwei:
@mkanoor @hsong-rh with new api-common gem released, can we move this PR on?

PR #302, opened 2020-03-17
Refactor RBAC to use pundit gem and permissions

hsong-rh:
This PR contains two changes on RBAC:
1. Refactor using Pundit gem;
2. Differentiate `admin` and `approver` by RBAC access permissions


hsong-rh:
@eclarizio 
Most time Approval uses `authorized` to do `resource level` permission check. It doesn't need the actual instance. The only exception is the second case you mentioned: an authorization in scope. That's the special case for parent request check when scoping child requests.

app/policies/mixins/rbac_mixin.rb
```diff
@@ -0,0 +1,102 @@
...
+      return true unless Insights::API::Common::RBAC::Access.enabled?
+
+      Insights::API::Common::RBAC::Access.new(klass.table_name, verb).process.accessible? ? true : (raise Exceptions::NotAuthorizedError, "#{verb.titleize} access not authorized for #{klass.table_name}")
+    end
+
+    # instance level check
+    def resource_instance_accessible?(resource, resource_id)
+      return true if admin?
+
+      approver? ? approvable?(resource, resource_id) : owned?(resource, resource_id)
```
bzwei: `approver? && approvable?() || requester? && owned?()`

app/controllers/application_controller.rb
```diff
@@ -40,4 +41,11 @@ def check_entitled(entitlement)
 
     required_entitlements.map { |e| entitlement.send(e) }.all?
   end
+
+  def pundit_user
+    UserContext.new(
+      Insights::API::Common::Request.current!,
+      params
```
mkanoor: @hsong-rh Do you need the params in your User context objects, In catalog we have one corner case where we need to check for the params for portfolio to copy to. If possible we should avoid params, since the policy scopes gets passed in the actual object. This also has problems once you go to user capabilities because those params won't be there when you are listing an object.

hsong-rh: Yes, I know `params` will not available for user capability. But Approval also has a corner case when we validate to create `action`,  we need to query `random_access_key` from its request defined in `params`.

app/controllers/api/v1x0/graphql_controller.rb
```diff
@@ -17,7 +17,8 @@ def overlay
         {
           "^.*$" => {
             "base_query" => lambda do |model_class, _ctx|
-              "::Api::V1x0::#{model_class.to_s.pluralize}Controller".constantize.new.send(:index_scope, model_class.all)
+              authorize model_class
+              model_class
```
mkanoor: @hsong-rh Should this be calling rbac_scope instead of authorize directly, Since you dont have the actual object to do the authorize on.
https://github.com/RedHatInsights/catalog-api/blob/0ae757293c7bfac2c1f02ddc49d8cc63e135245f/app/controllers/api/v1/graphql_controller.rb#L12

app/controllers/api/v1x0/requests_controller.rb
```diff
@@ -1,64 +1,34 @@
...
       def requests_prefilter
-        return nil unless params[:request_id]
+        return Request unless params[:request_id]
 
-        resource_check('read', params[:request_id], Request) # NotAuthorizedError if current user cannot access parent request
+        authorize Request.find(params.require(:request_id))
         Request.find(params[:request_id]).children
       end
-
-      def rbac_scope(relation)
```
mkanoor: @hsong-rh I think we might still need the rbac scope which can be called for index calls. This rbac_sope should also be called from the graphql overlay.

app/controllers/api/v1x0/workflows_controller.rb
```diff
@@ -1,23 +1,20 @@
...
       def create
+        authorize Workflow
+
         workflow = WorkflowCreateService.new(params.require(:template_id)).create(params_for_create)
         json_response(workflow, :created)
       end
 
-      # TODO: remove 'approval:workflows:read' from approver acls list in RBAC Insight
       def show
-        raise Exceptions::NotAuthorizedError, "Not Authorized for workflows" if Insights::API::Common::RBAC::Access.enabled? && !admin?
```
mkanoor: @hsong-rh This is a good change, are you removing all RBAC enabled check and assuming every spec is going to be rbac enabled like it should be.

app/policies/action_policy.rb
```diff
@@ -0,0 +1,45 @@
+class ActionPolicy < ApplicationPolicy
+
+  ADMIN_OPERATIONS     = [Action::MEMO_OPERATION, Action::APPROVE_OPERATION, Action::DENY_OPERATION, Action::CANCEL_OPERATION].freeze
+  APPROVER_OPERATIONS  = [Action::MEMO_OPERATION, Action::APPROVE_OPERATION, Action::DENY_OPERATION].freeze
+  REQUESTER_OPERATIONS = [Action::CANCEL_OPERATION].freeze
+
+  class Scope < ApplicationPolicy::Scope
+    def resolve
```
mkanoor: @hsong-rh How about this

```
def resolve
    my_scopes = @user.access_object.scopes('action','read')
    if my_scopes.include?('admin')
       scope.all
    elsif my_scopes.include?('group')
       action_ids = approver_id_list(scope.model.table_name)
       Rails.logger.debug { "Approver scope for actions: #{action_ids}" }
       scope.where(:id => action_ids)
    elsif my_scopes.include?('user')
       # What should the user see is that by owner
    else
       raise Exceptions::NotAuthorizedError, "Read access not authorized for #{scope}"
    end
end
```

app/policies/user_context.rb
```diff
@@ -0,0 +1,8 @@
+class UserContext
+  attr_reader :request, :params
+
+  def initialize(request, params)
+    @request = request
+    @params = params
+  end
```
mkanoor: @hsong-rh You might want to move the access object here

```
attr_reader :access_object

def access_object
    @access_object ||= Insights::API::Common::RBAC::Access.new.process
end

```

By doing it this way you will have access to the access_object in the resolve method of ApplicationPolicy and its sub classes since they all have the user object.

app/policies/mixins/rbac_mixin.rb
```diff
@@ -0,0 +1,113 @@
...
+      Request.where(:group_ref => assigned_group_refs, :state => APPROVER_VISIBLE_STATES).pluck(:id)
+    end
+
+    def assigned_group_refs
+      Insights::API::Common::RBAC::Service.call(RBACApiClient::GroupApi) do |api|
+        Insights::API::Common::RBAC::Service.paginate(api, :list_groups, :scope => 'principal').collect(&:uuid)
+      end
+    end
+
+    def access
```
mkanoor: @hsong-rh It might be better to move this to the UserContext class so it can be accessed from the resolve method in the Policy objects.

app/controllers/api/v1x0/graphql_controller.rb
mkanoor: @hsong-rh These are all index calls with read permission. The admin, approver and user could all make this call.

hsong-rh: `authorize ModelClass` is a class level permission check, not necessary on a certain instance. Here we just need the `read` permission check on `ModelClass` level

app/controllers/api/v1x0/requests_controller.rb
hsong-rh: No, the logic of `rbac_scope` is moved to `resolve` method, called by `policy_scope`.

app/controllers/api/v1x0/workflows_controller.rb
hsong-rh: I moved the check to https://github.com/RedHatInsights/approval-api/blob/a1806ddd501e19ed2dca133799dd1092f826c38a/app/policies/mixins/rbac_mixin.rb#L14. It will be always called there.

app/policies/user_context.rb
hsong-rh: 👍 I will change this

app/policies/action_policy.rb
hsong-rh: 👍

app/policies/mixins/rbac_mixin.rb
hsong-rh: Ok, will change it.

app/controllers/api/v1x0/graphql_controller.rb
mkanoor: @hsong-rh This is a sample GraphQL call note the ID inside

```
{
  requests(filter: {parent_id: "360"}) {
    id
    actions {
      id
      operation
      comments
      created_at
      processed_by
    }
    number_of_children
    decision
    description
    number_of_finished_children
    parent_id
  }
}
```
If you just check for the Class level access how would you prevent users from passing in any  parent_id and downloading the model.

app/controllers/api/v1x0/actions_controller.rb
```diff
@@ -1,61 +1,31 @@
...
-      APPROVER_OPERATIONS  = [Action::MEMO_OPERATION, Action::APPROVE_OPERATION, Action::DENY_OPERATION].freeze
-      REQUESTER_OPERATIONS = [Action::CANCEL_OPERATION].freeze
-
-      before_action :read_access_check, :only => %i[show]
-      before_action :create_access_check, :validate_create_action, :only => %i[create]
+      include Mixins::IndexMixin
 
       def index
         req = Request.find(params.require(:request_id))
+        authorize(req, :policy_class => RequestPolicy)
```
bzwei: why do you need to `authorize` and `policy_scope` together? shouldn't `policy_scope` already do permission check?

app/policies/request_policy.rb
```diff
@@ -0,0 +1,39 @@
...
+
+      # parent request index
+      if scope.class == Class
+        return scope.where(:parent_id => nil) if admin?(scope)
+        return scope.where(:id => approver_id_list(scope.table_name)) if approver?(scope)
+
+        scope.where(:parent_id => nil, :id => owner_id_list(scope.table_name))
+      # child request index
+      else
+        return scope if admin?(scope.model)
```
bzwei: We may be able to simply `return scope`. Approver never be able to reach here because he will be rejected at parent level. Requester should be able to see all children if he passes the parent id check.

app/controllers/api/v1x0/actions_controller.rb
hsong-rh: `authorize` checks on `Request`, `policy_scope` checks on `Action` collections

app/controllers/api/v1x0/graphql_controller.rb
hsong-rh: @mkanoor 👍 Will work on this kind of user case.

app/policies/mixins/rbac_mixin.rb
```diff
@@ -0,0 +1,113 @@
...
+
+    # instance level check
+    def resource_instance_accessible?(resource, resource_id)
+      return true if admin?
+
+      approver? && approvable?(resource, resource_id) || requester? && owned?(resource, resource_id)
+    end
+
+    def admin?(resource = @record)
+      return false unless Insights::API::Common::RBAC::Access.enabled?
```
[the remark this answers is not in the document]
hsong-rh: Will remove the `enabled?` checking

app/policies/mixins/rbac_mixin.rb
```diff
@@ -0,0 +1,113 @@
...
+
+      klass = resource_class(resource)
+      access.admin_scope?(klass.table_name, 'read', APP_NAME)
```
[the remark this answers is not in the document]
hsong-rh: Both `admin_scope?` and `approver_scope?` need the second `verb` arguments. In Approval only `show` and `index` need to check the scope. They all corresponds to `read` action

app/policies/mixins/rbac_mixin.rb
```diff
@@ -0,0 +1,113 @@
...
+
+    def approver?(resource = @record)
+      return false unless Insights::API::Common::RBAC::Access.enabled?
+
+      klass = resource_class(resource)
+      access.group_scope?(klass.table_name, 'read', APP_NAME)
+    end
+
+    def requester?(resource = @record)
+      !admin?(resource) && !approver?(resource)
```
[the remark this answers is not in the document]
hsong-rh: 👍

app/controllers/api/v1x0/requests_controller.rb
```diff
@@ -1,64 +1,34 @@
...
       end
 
       private
 
       def requests_prefilter
-        return nil unless params[:request_id]
+        return Request unless params[:request_id]
 
-        resource_check('read', params[:request_id], Request) # NotAuthorizedError if current user cannot access parent request
+        authorize Request.find(params.require(:request_id))
```
bzwei: curious how does pundit know which query method to use for `authorize`.

app/controllers/api/v1x0/templates_controller.rb
```diff
@@ -1,24 +1,15 @@
...
     class TemplatesController < ApplicationController
-      include Api::V1x0::Mixins::IndexMixin
-      include Api::V1x0::Mixins::RBACMixin
-
-      before_action :read_access_check, :only => %i[show]
+      include Mixins::IndexMixin
 
       def show
-        template = Template.find(params.require(:id))
+        template = policy_scope(Template).find(params.require(:id))
```
bzwei: why can't we find the template and `authorize` like in other controller?

app/controllers/api/v1x0/requests_controller.rb
hsong-rh: Based on the method in controller class which calls `authorized`, by default pundit will find the corresponding `policy` class and call the method whose name is appending with `?`. Can also use parameters to change `policy` and `query` method, like here.

app/controllers/api/v1x0/templates_controller.rb
hsong-rh: Yes, we can, but hey are the same for `template`

app/policies/action_policy.rb
hsong-rh: @mkanoor I prefer to put class level `permission_check` in the beginning. There's no need to proceed further if it fails.

mkanoor: @hsong-rh permission_check is not using the scopes. Its just checking if you have the permission. The scopes method allows you do the 2 things
(1) Check permission
(2) Get Scope
Versus
scopes in a single pass collect all the scopes. If the scopes result is empty you dont have access

app/policies/mixins/rbac_mixin.rb
```diff
@@ -0,0 +1,106 @@
...
+      Request.where(:group_ref => assigned_group_refs, :state => APPROVER_VISIBLE_STATES).pluck(:id)
+    end
+
+    def assigned_group_refs
+      Insights::API::Common::RBAC::Service.call(RBACApiClient::GroupApi) do |api|
+        Insights::API::Common::RBAC::Service.paginate(api, :list_groups, :scope => 'principal').collect(&:uuid)
+      end
+    end
+
+    def resource_class(resource)
```
mkanoor: @hsong-rh This seems overly complex
https://github.com/RedHatInsights/catalog-api/blob/358797e376b284f9cdeab6038a1fcafcf9cb3287/lib/catalog/rbac/access.rb#L43

app/controllers/api/v1x0/actions_controller.rb
```diff
@@ -1,61 +1,31 @@
...
-        return relation if admin?
-
-        # Only approver can reach here
-        resource_check('read', params[:request_id], Request) # NotAuthorizedError if current user cannot access the particular request
-
-        action_ids = approver_id_list(relation.model.table_name)
-        Rails.logger.debug { "Approver scope for actions: #{action_ids}" }
+        # make sure have permission to read the request
+        req = Request.find(params.require(:request_id))
+        authorize(req, 'show?', :policy_class => RequestPolicy)
```
mkanoor: @hsong-rh Why are we passing in the RequestPolicy all the time, can't we figure it out from the parent object. In Catalog we have Portfolio Item which is child of Portfolio. So in the Policy for PortfolioItem we just pick up the Portfolio.
https://github.com/RedHatInsights/catalog-api/blob/358797e376b284f9cdeab6038a1fcafcf9cb3287/app/policies/portfolio_item_policy.rb#L8

app/policies/mixins/rbac_mixin.rb
hsong-rh: The reason we defined this is because it's shared by all different caller: `permission_check`, 'admin?`, `approver?` and `requester?`. Different callers may send different type of params. Here in one place we handle all scenarios.

app/policies/action_policy.rb
hsong-rh: Yes, I know the difference. My point is put `permission_check` in the beginning. If it fails, there's no need to continue. `scope` check is the second step, like what we define here.

app/controllers/api/v1x0/requests_controller.rb
```diff
@@ -1,64 +1,30 @@
...
-          relation = Request.all unless relation
-          ids = approver_id_list(relation.model.table_name)
-        when PERSONA_REQUESTER, nil
-          relation = Request.where(:parent_id => nil) unless relation
-
-          ids = owner_id_list(relation.model.table_name)
-        else
-          raise Exceptions::NotAuthorizedError, "Unknown persona"
-        end
+        return collection(policy_scope(Request)) unless params[:request_id]
```
mkanoor: @hsong-rh This is a bit confusing, couldn't we create a relation that looked like
relation =params[:request_id) ? Request.where(:id => params[:request_id)) : Request.all

app/controllers/api/v1x0/actions_controller.rb
hsong-rh: Catalog did the same thing, from child to pick up parent, right? Catalog did the similar logic in child policy. Approval checks and may stop it in parent level.

app/controllers/api/v1x0/requests_controller.rb
hsong-rh: Not that simple. We still need an `authorize` check on parent request if `params[:request_id]` exists

app/controllers/api/v1x0/actions_controller.rb
mkanoor: @hsong-rh We have Policies for each of the objects and do the appropriate thing from that Policy object.
e.g. https://github.com/RedHatInsights/catalog-api/tree/master/app/policies
The Policy object does the right thing it knows how to do the validation by looking at the appropriate parent object. This allows us to move the logic from the controller to the Policy objects and allows us to change the logic in the Policy object in the future.

app/policies/action_policy.rb
mkanoor: @hsong-rh So what happens if you have access, then you call the function again it has to loop thru all the permission one more time to pick up the scope. You are optimizing for what happens if there is no access, but if there is access you have to traverse the list again to collect scopes.

app/controllers/api/v1x0/requests_controller.rb
mkanoor: @hsong-rh What api end point is this

/api/approval/requests.  (This is an index call)
or
/api/approval/requests/555 (This is a show call)

If there is filter involved where id = 555 on the index call that is handled separately by the filtering code.

app/policies/action_policy.rb
```diff
@@ -0,0 +1,45 @@
...
+      raise Exceptions::NotAuthorizedError, "Read access not authorized for #{scope}" unless permission_check('read', scope)
+      return scope.all if admin?(scope.model)
+
+      action_ids = approver_id_list(scope.model.table_name)
+      Rails.logger.debug { "Approver scope for actions: #{action_ids}" }
```
mkanoor: @hsong-rh This is for both user and approver correct or is this only for approver?

app/policies/action_policy.rb
hsong-rh: No, `permission_check` only runs one time and it only checks permission, never list resources.

app/policies/action_policy.rb
```diff
@@ -0,0 +1,45 @@
...
+  def create?
+    permission_check('create')
+    validate_create_action
+  end
+
+  def show?
+    resource_check('read')
+  end
+
+  def query?
```
mkanoor: @hsong-rh Who calls this method, why is it different from read.

app/controllers/api/v1x0/actions_controller.rb
hsong-rh: Ok, I will work to move the logic to its `policy` class.

app/policies/action_policy.rb
hsong-rh: So far it's only for `approver`, because `user` now doesn't have `approval:actions:read` permission yet.

app/controllers/api/v1x0/requests_controller.rb
hsong-rh: Approval has another API for index call on parent request id, e.g. `/api/approval/requests/555/requests`. This's the reason why we need a separate `authorize` on parent request first

app/policies/mixins/rbac_mixin.rb
```diff
@@ -0,0 +1,106 @@
+module Mixins
+  module RBACMixin
```
eclarizio: The biggest problem I have with this being a mixin is that this contains *a lot* of logic. When you mix this into another class, you're adding all of that functionality and that makes it really difficult to test.

I know we're trying to get this PR through rather quickly, so it should probably be handled in a separate PR, but I would like to see this mixin broken out into a separate class and then a corresponding spec written against it. I would imagine you'll find it pretty ridiculous the levels of contexts you would need to create (ex: #resource check has easily 10+ levels of cyclomatic complexity). This would further indicate that we could potentially break it up into smaller pieces and have simpler classes to follow.

hsong-rh: Refactored per discussion with Madhu. The left logic is necessary to be shared with different policys.

app/controllers/application_controller.rb
```diff
@@ -40,4 +41,8 @@ def check_entitled(entitlement)
 
     required_entitlements.map { |e| entitlement.send(e) }.all?
   end
+
+  def pundit_user
+    UserContext.new(Insights::API::Common::Request.current!, params)
+  end
```
mkanoor: @hsong-rh I think this should be cached
  ```
   def pundit_user
    @user ||= UserContext.new(Insights::API::Common::Request.current!, params)
   end
```

File: app/controllers/api/v1x0/templates_controller.rb [hunk 23]
```diff
@@ -1,24 +1,17 @@
...
     class TemplatesController < ApplicationController
-      include Api::V1x0::Mixins::IndexMixin
-      include Api::V1x0::Mixins::RBACMixin
-
-      before_action :read_access_check, :only => %i[show]
+      include Mixins::IndexMixin
 
       def show
         template = Template.find(params.require(:id))
+        authorize template
```
mkanoor: @hsong-rh Should there be permissions for Template in the approval.json. I didn't see it in the rbac_config PR.
Is the template the same as workflow?

hsong-rh: We don have permission for template in approval.json. Only have `approval:templates:read` for approval admin

app/controllers/application_controller.rb
hsong-rh: Why?  `params` may be different

hsong-rh: Changed.

bzwei:
@eclarizio @mkanoor please approve if you are ok with the PR.

PR #308, opened 2020-03-24
API /requests lists child requests visible to approver

bzwei:
https://projects.engineering.redhat.com/browse/SSP-1370

Before for endpoint `/requests` the controller always limits the collection to parent requests where `:parent_id => nil` and continues to narrow the collection according to persona. However approvers can see never see parent requests but only qualified children.

With this fix the endpoint now includes child requests that approvers can see.

Also we discover that `access_control_entries` are no longer needed because each child request has `group_ref` attribute to filter approvers and `owner` attribute to filter requesters. 

A follow-up PR will completely remove `access_control_entries` table.

### DIFF
--- a/app/controllers/api/v1x0/mixins/rbac_mixin.rb
+++ b/app/controllers/api/v1x0/mixins/rbac_mixin.rb
@@ -6,6 +6,7 @@ module Api
 
         ADMIN_ROLE = 'Approval Administrator'.freeze
         APPROVER_ROLE = 'Approval Approver'.freeze
+        APPROVER_VISIBLE_STATES = [ApprovalStates::NOTIFIED_STATE, ApprovalStates::COMPLETED_STATE].freeze
 
         # create action only needs class level permission check
         def create_access_check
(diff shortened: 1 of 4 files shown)


app/controllers/api/v1x0/mixins/rbac_mixin.rb
```diff
@@ -130,19 +130,14 @@ def owner_request_ids
         # All child request ids for approver to process
         def visible_request_ids_for_approver
           visible_states = [ApprovalStates::NOTIFIED_STATE, ApprovalStates::COMPLETED_STATE]
-          Request.where(:workflow_id => workflow_ids, :state => visible_states).pluck(:id)
+          Request.where(:group_ref => assigned_group_refs, :state => visible_states).pluck(:id)
```
gmcculloug: This looks good, just wanted to comment that `visible_states` should be a constant, either defined here or in approval_states.rb.

app/controllers/api/v1x0/requests_controller.rb
```diff
@@ -26,7 +26,7 @@ def index
                      resource_check('read', params[:request_id], Request) # NotAuthorizedError if current user cannot access parent request
                      Request.find(params[:request_id]).children
                    else
-                     Request.where(:parent_id => nil)
+                     nil
```
gmcculloug: This would be more of a refactoring, but would be nice if `requests` became a private method then the `index` method calls `requests` with `params`.

The logic could then be written to return early if `request_id` is nil:
```ruby
def requests(params)
  return nil if params[:request_id].blank?

  resource_check('read', params[:request_id], Request) # NotAuthorizedError if current user cannot access parent request
  Request.find(params[:request_id]).children
end
```

hsong-rh: 👍

app/controllers/api/v1x0/requests_controller.rb
```diff
@@ -22,32 +22,37 @@ def show
...
-            raise Exceptions::NotAuthorizedError, "No permission to access requests assigned to approvers" unless approver?
-
-            approver_id_list(relation.model.table_name)
-          when PERSONA_REQUESTER, nil
-            owner_id_list(relation.model.table_name)
-          else
-            raise Exceptions::NotAuthorizedError, "Unknown persona"
-          end
+        ids = nil
+        case Insights::API::Common::Request.current.headers[Insights::API::Common::Request::PERSONA_KEY]
```
mkanoor: @bzwei You had mentioned that you are going to be removing the personas, so do we still want to implement a solution with the personas?

bzwei: Yes. We are going to remove personas once capabilities are introduced. For now we still need to maintain the code as is.

PR #310, opened 2020-03-26
Store name together with uuid for group_refs in workflows

bzwei:
https://projects.engineering.redhat.com/browse/SSP-1354

Format of group_refs changed in workflow API, therefore made a new version v1.2.

### DIFF
--- a/app/services/workflow_get_service.rb
+++ b/app/services/workflow_get_service.rb
@@ -1,0 +1,16 @@
+require_relative 'mixins/group_validate_mixin'
+
+class WorkflowGetService
+  include GroupValidateMixin
+  attr_accessor :workflow_id
+
+  def initialize(workflow_id)
+    self.workflow_id = workflow_id
+  end
+
+  def get
+    Workflow.find(workflow_id).tap do |workflow|
+      validate_and_update_approver_groups(workflow, false)
+    end
+  end
+end
--- a/app/services/workflow_link_service.rb
+++ b/app/services/workflow_link_service.rb
@@ -11,7 +11,7 @@ class WorkflowLinkService
   end
 
   def link(tag_attrs)
-    validate_approver_groups(Workflow.find(workflow_id).group_refs)
+    validate_and_update_approver_groups(Workflow.find(workflow_id))
 
     TagLink.find_or_create_by!(tag_link(tag_attrs))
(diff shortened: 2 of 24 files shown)


app/services/workflow_link_service.rb
```diff
@@ -11,7 +11,7 @@ def initialize(workflow_id)
   end
 
   def link(tag_attrs)
-    validate_approver_groups(Workflow.find(workflow_id).group_refs)
+    validate_and_update_approver_groups(Workflow.find(workflow_id))
```
hsong-rh: Why do we need to update `workflow` here? We only validate before, otherwise raise an exception

app/services/workflow_update_service.rb
```diff
@@ -1,15 +1,17 @@
...
     self.workflow = Workflow.find(workflow_id)
   end
 
   def update(options)
-    options[:access_control_entries] =
-      (options[:group_refs] || []).collect do |uuid|
-        AccessControlEntry.new(:group_uuid => uuid, :permission => 'approve')
-      end
+    if options[:group_refs]
+      options[:group_refs] = validate_approver_groups(options[:group_refs])
```
hsong-rh: Should we call `validate_and_update_approver_groups` here? so don't need line 16 anymore

app/services/workflow_get_service.rb
```diff
@@ -0,0 +1,16 @@
...
+
+class WorkflowGetService
+  include GroupValidateMixin
+  attr_accessor :workflow_id
+
+  def initialize(workflow_id)
+    self.workflow_id = workflow_id
+  end
+
+  def get
```
hsong-rh: Why we need `validate_and_update_approver_groups` for each `get`? Just want to update the existing `group_refs` attribute to new format?

app/services/workflow_link_service.rb
bzwei: It gives us a chance to refresh the group name which is now also stored in our DB.

app/services/workflow_update_service.rb
bzwei: no. the workflow is saved in the next line. We don't need to save multiple times.

app/services/workflow_get_service.rb
bzwei: To refresh group name which may be out of sync from rbac.

PR #313, opened 2020-04-01
Enabled user capabilities in Approval

hsong-rh:
Added user capabilities feature in Approval. The user_capabilities in metadata will have following keys. Values can be either true or false:

1. Template: `"metadata"=>{"user_capabilities"=>{"show"=>true}}`
2. Workflow: `"metadata"=>{"user_capabilities"=>{"update"=>true, "destroy"=>true, "link"=>true, "unlink"=>true, "create"=>true, "show"=>true}}`
3. Request:  `"metadata"=>{"user_capabilities"=>{"approve"=>true, "cancel"=>true, "create"=>true, "deny"=>true, "memo"=>true, "show"=>true}}`

Also having refactoring, moved `authorize` check logic in `index` to corresponding `policy scope`.

### DIFF
--- a/app/controllers/api/v1x0/actions_controller.rb
+++ b/app/controllers/api/v1x0/actions_controller.rb
@@ -5,8 +5,6 @@ module Api
 
       def index
         req = Request.find(params.require(:request_id))
-        authorize req, :show?
-
         collection(policy_scope(req.actions))
       end
 
--- a/app/controllers/api/v1x0/actions_controller.rb
+++ b/app/controllers/api/v1x0/actions_controller.rb
@@ -4,8 +4,7 @@ module Api
       include Mixins::IndexMixin
 
       def index
-        req = Request.find(params.require(:request_id))
-        collection(policy_scope(req.actions))
+        collection(policy_scope(Action))
       end
 
       def show
--- a/app/controllers/api/v1x0/graphql_controller.rb
+++ b/app/controllers/api/v1x0/graphql_controller.rb
@@ -17,7 +17,7 @@ module Api
         {
           "^.*$" => {
             "base_query" => lambda do |model_class, _ctx|
-              policy_scope(model_class.all)
+              policy_scope(model_class)
             end
           }
         }
--- a/app/controllers/api/v1x0/graphql_controller.rb
+++ b/app/controllers/api/v1x0/graphql_controller.rb
@@ -17,7 +17,7 @@ module Api
         {
           "^.*$" => {
             "base_query" => lambda do |model_class, _ctx|
-              policy_scope(model_class)
+              policy_scope(model_class.all)
             end
           }
         }
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -17,19 +17,17 @@ class ApplicationController < ActionController::API
     Insights::API::Common::Request.with_request(request) do |current|
       Rails.logger.info("Request started #{request.original_url}")
 
-      if current.required_auth?
-        raise Insights::API::Common::EntitlementError, "User not Entitled" unless check_entitled(current.entitlement)
+      UserContext.with_user_context(UserContext.new(current, params)) do |user_context|
+        if current.required_auth?
+          raise Insights::API::Common::EntitlementError, "User not Entitled" unless check_entitled(current.entitlement)
 
-        Thread.current[:user] = UserContext.new(request, params)
-        ActsAsTenant.with_tenant(current_tenant(current)) { yield.tap { Rails.logger.info("Request ended #{request.original_url}") } }
-      else
-        ActsAsTenant.without_tenant { yield.tap {Rails.logger.info("Request ended #{request.original_url}")} }
+          ActsAsTenant.with_tenant(current_tenant(current)) { yield.tap { Rails.logger.info("Request ended #{request.original_url}") } }
+        else
+          ActsAsTenant.without_tenant { yield.tap {Rails.logger.info("Request ended #{request.original_url}")} }
+        end
       end
-
     rescue Exceptions::NoTenantError
       json_response({:message => 'Unauthorized'}, :unauthorized)
-    ensure
-      Thread.current[:user] = nil
     end
   end
 
@@ -46,6 +44,6 @@ class ApplicationController < ActionController::API
   end
 
   def pundit_user
-    Thread.current[:user]
+    UserContext.current_user_context
   end
 end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -20,6 +20,7 @@ class ApplicationController < ActionController::API
       if current.required_auth?
         raise Insights::API::Common::EntitlementError, "User not Entitled" unless check_entitled(current.entitlement)
 
+        Thread.current[:user] = UserContext.new(request, params)
         ActsAsTenant.with_tenant(current_tenant(current)) { yield.tap { Rails.logger.info("Request ended #{request.original_url}") } }
       else
         ActsAsTenant.without_tenant { yield.tap {Rails.logger.info("Request ended #{request.original_url}")} }
@@ -27,6 +28,8 @@ class ApplicationController < ActionController::API
 
     rescue Exceptions::NoTenantError
       json_response({:message => 'Unauthorized'}, :unauthorized)
+    ensure
+      Thread.current[:user] = nil
     end
   end
 
@@ -43,6 +46,6 @@ class ApplicationController < ActionController::API
   end
 
   def pundit_user
-    @user ||= UserContext.new(Insights::API::Common::Request.current!, params)
+    Thread.current[:user]
   end
 end
--- a/app/models/application_record.rb
+++ b/app/models/application_record.rb
@@ -9,6 +9,6 @@ class ApplicationRecord < ActiveRecord::Base
   end
 
   def user_context
-    @user_context ||= UserContext.new(Insights::API::Common::Request.current!)
+    Thread.current[:user]
   end
 end
--- a/app/models/application_record.rb
+++ b/app/models/application_record.rb
@@ -1,14 +1,19 @@
 require "acts_as_tenant"
 
 class ApplicationRecord < ActiveRecord::Base
+  include Pundit
+
   self.abstract_class = true
-  attribute :metadata, ActiveRecord::Type::Json.new
 
   def metadata
-    { :user_capabilities => {} }
+    user_context.nil? ? { :user_capabilities => {} } : {:user_capabilities => policy_name.new(user_context, self).user_capabilities}
   end
 
   def user_context
     Thread.current[:user]
+  end
+
+  def policy_name
+    PolicyFinder.new(self).policy
   end
 end
--- a/app/models/application_record.rb
+++ b/app/models/application_record.rb
@@ -5,15 +5,7 @@ class ApplicationRecord < ActiveRecord::Base
 
   self.abstract_class = true
 
-  def metadata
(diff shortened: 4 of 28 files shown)


app/controllers/api/v1x0/graphql_controller.rb
```diff
@@ -17,7 +17,7 @@ def overlay
         {
           "^.*$" => {
             "base_query" => lambda do |model_class, _ctx|
-              policy_scope(model_class)
+              policy_scope(model_class.all)
```
mkanoor, line 20: @hsong-rh What prompted this change? policy_scope takes in a relationship and not a class?

bzwei: class should be sufficient and preferred.

hsong-rh: Both relation and class are ok for policy_scope. The change here just keeps in sync with other places.

app/policies/action_policy.rb
```diff
@@ -1,13 +1,13 @@
...
-
-  ADMIN_OPERATIONS     = [Action::MEMO_OPERATION, Action::APPROVE_OPERATION, Action::DENY_OPERATION, Action::CANCEL_OPERATION].freeze
-  APPROVER_OPERATIONS  = [Action::MEMO_OPERATION, Action::APPROVE_OPERATION, Action::DENY_OPERATION].freeze
-  REQUESTER_OPERATIONS = [Action::CANCEL_OPERATION].freeze
-
   class Scope < ApplicationPolicy::Scope
     def resolve
       raise Exceptions::NotAuthorizedError, "Read access not authorized for #{scope}" unless permission_check('read', scope.model)
 
+      if user.params[:request_id]
```
mkanoor, line 6: @hsong-rh If user.params is nil that if statement might fail. Most of the time for index call you will have params from the Controller. You might want to raise an Arguments error if its missing.

app/policies/application_policy.rb
```diff
@@ -36,6 +36,20 @@ def destroy?
     false
   end
 
+  def query?
```
mkanoor, line 39: @hsong-rh Can't this be show?

hsong-rh: This method is for `graphql` only. It will not display in `metadata`

app/controllers/api/v1x0/graphql_controller.rb
hsong-rh: Changed to use class

app/policies/application_policy.rb
hsong-rh: Removed

app/models/application_record.rb
```diff
@@ -2,4 +2,13 @@
 
 class ApplicationRecord < ActiveRecord::Base
   self.abstract_class = true
+  attribute :metadata, ActiveRecord::Type::Json.new
+
+  def metadata
+    { :user_capabilities => {} }
+  end
+
+  def user_context
```
mkanoor, line 11: @hsong-rh This is going to be done for every instance, if you have 50 requests we will create this object 50 times each one making a direct call to RBAC slowing things do. We should try caching it one per session in the 
https://github.com/RedHatInsights/approval-api/blob/7248da7d6a8182cbc1ba73f651d8ff1035093c24/app/controllers/application_controller.rb#L22
With an ensure to clear out the cached value.

app/policies/request_policy.rb
```diff
@@ -32,13 +37,43 @@ def show?
...
+              elsif approver?(record.class)
+                Action::APPROVER_OPERATIONS & valid_actions_on_state
+              else
+                Action::REQUESTER_OPERATIONS & valid_actions_on_state
+              end
+
+    # only child request can be approved/denied
+    actions -= [Action::APPROVE_OPERATION, Action::DENY_OPERATION] if record.parent?
+    # only root can be cancelled
+    actions -= [Action::CANCEL_OPERATION] unless record.root?
```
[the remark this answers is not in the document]
hsong-rh: After line 50 `actions` has been filtered by permission scope

app/controllers/application_controller.rb
```diff
@@ -43,6 +45,6 @@ def check_entitled(entitlement)
   end
 
   def pundit_user
-    @user ||= UserContext.new(Insights::API::Common::Request.current!, params)
+    @user ||= Thread.current[:user]
```
mkanoor, line 48: @hsong-rh This should always be Thread.current[:user]

app/policies/request_policy.rb
```diff
@@ -32,13 +37,43 @@ def show?
...
+    super.merge(valid_actions_hash)
+  end
+
+  def valid_actions_hash
+    hash = { Action::APPROVE_OPERATION => false,
+             Action::CANCEL_OPERATION  => false,
+             Action::DENY_OPERATION    => false,
+             Action::MEMO_OPERATION    => true }
+
+    actions = if admin?(record.class)
```
bzwei, line 50: Which api will return capabilities - get, list, or both? Here you only consider the login user has single role, but he can be a combination of all roles. For list the role is selected by persona header; for get we probably need to combine all actual roles.

mkanoor: @bzwei Its for list and show calls, anytime we fetch an object these attributes will be added. The object definition has been updated in the openapi.json

hsong-rh: @bzwei I think the user's capabilities shouldn't set by checking persona. I will add the logic to deal with multiple roles combination

app/controllers/application_controller.rb
```diff
@@ -43,6 +44,6 @@ def check_entitled(entitlement)
   end
 
   def pundit_user
-    @user ||= UserContext.new(Insights::API::Common::Request.current!, params)
+    Thread.current[:user_context]
```
mkanoor, line 47: @hsong-rh This should be
```
UserContext.current_user_context
```

app/models/application_record.rb
```diff
@@ -1,5 +1,11 @@
 require "acts_as_tenant"
 
 class ApplicationRecord < ActiveRecord::Base
+  include Pundit
+
   self.abstract_class = true
+
+  def user_context
+    Thread.current[:user_context]
```
mkanoor, line 9: @hsong-rh This should be
```
UserContext.current_user_context
```

app/models/concerns/metadata.rb
```diff
@@ -0,0 +1,11 @@
+module Metadata
+  extend ActiveSupport::Concern
+
+  included do
+    attribute :metadata, ActiveRecord::Type::Json.new
+
+    def metadata
```
mkanoor, line 7: @hsong-rh 
One issue we have here is we need to be able to add extra data into the metadata other than just the user_capabilities. For example in catalog we add extra properties like 
'shared' => true|false

If we have to do this then we have to override the metadata method in the included class name.

https://github.com/RedHatInsights/catalog-api/blob/40e8e8749f5aef59abb0864e4e1acfb8d8549544/app/models/portfolio.rb#L24

Should the mixin be called user_capabilities instead of metadata?

app/models/concerns/metadata.rb
```diff
@@ -0,0 +1,15 @@
...
+      { :user_capabilities => user_capabilities, :shared => true }
```
mkanoor, line 8: @hsong-rh can you remove the :shared => true that was just an example

app/controllers/api/v1x0/actions_controller.rb
```diff
@@ -5,8 +5,6 @@ class ActionsController < ApplicationController
 
       def index
         req = Request.find(params.require(:request_id))
-        authorize req, :show?
```
bzwei, line 8: I feel leaving `authorize req`  here is better than moving it inside the `policy_scope`

mkanoor: @bzwei When GraphQL makes a request the logic is similar to an index call, so its better to do the check in policy_scope in one place. From the pundit gems perspective all the index calls are scope based and that's where the check should happen for individual calls like show/update/edit/delete which work on a single object you can use authorize.

bzwei: GraphQL only does index from a model class, it does not pass has_many relation to the policy_scope. Look at the block https://github.com/RedHatInsights/approval-api/pull/313/files#diff-c423c5c790991b9cc7a73787c2cbe576R7, there we need to call `Request.find(params.require(:request_id))` again. It also checks permission and raise error which is already done by `authorize req`. We are re-inventing the logic but not utilizing the benefit from pundit.
Here the only line `authorize req, :show?` is simple and strait forward, easy to understand. GraphQL does not need this line anyway.

app/models/application_record.rb
```diff
@@ -1,5 +1,11 @@
 require "acts_as_tenant"
 
 class ApplicationRecord < ActiveRecord::Base
+  include Pundit
```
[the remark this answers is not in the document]
hsong-rh: Leftover when using `PolicyFinder`. Removed

app/policies/request_policy.rb
```diff
@@ -6,19 +6,27 @@ class Scope < ApplicationPolicy::Scope
...
-        raise Exceptions::NotAuthorizedError, "No permission to access requests assigned to approvers" unless approver?(klass)
-        approver_visible_requests(scope)
-      when PERSONA_REQUESTER, nil
-        scope == Request ? scope.by_owner.where(:parent_id => nil) : scope
+
+      if user.params[:request_id]
+        req = Request.find(user.params[:request_id])
+        raise Exceptions::NotAuthorizedError, "Read access not authorized for request #{req.id}" unless resource_check('read', req)
+        req.children
+      # GraphQL comes in with model_class.all, whose scope is respond to model call
```
mkanoor, line 14: @hsong-rh Can't we fix this here
https://github.com/RedHatInsights/approval-api/blob/830f34c87db22b513b70a301f1907e4d80b29270/app/controllers/api/v1x0/graphql_controller.rb#L20

Always call model

app/models/request.rb
```diff
@@ -83,4 +84,8 @@ def set_defaults
     self.number_of_children = 0
     self.number_of_finished_children = 0
   end
+
+  def policy_name
```
bzwei, line 88: Name implies string. Change to `policy_class`. We may be able to move this method to metadata.rb
```
def policy_class
  "#{model_name.name}Policy".contantize
end
```
```

File: app/policies/request_policy.rb (hunk 14 above)
Comment (bzwei) [reply app/policies/request_policy.rb:14]: @mkanoor we use model or scope to distinguish calling from regular controller or GraphQL. The result is different.

File: app/policies/request_policy.rb [hunk 16]
```diff
@@ -32,8 +40,63 @@ def show?
...
+      Request::CANCELED_STATE  => [Action::MEMO_OPERATION]
+    }[state]
+  end
+
+  def valide_actions_on_roles
+    admin     = admin?(record.class)
+    approver  = approver?(record.class)
+    requester = requester?(record.class)
+
+    if admin && approver && requester
```
bzwei, line 84: can we do something like this
```
  operations = []
  operations |= Actions::ADMIN_PERATIONS if admin?(record.class)
  operations |= Actions::APPROVER_OPERATIONS if approver?(record.class)
  operations |= Actions::REQUESTER_OPERATIONS if requester?(record.class)

  operations
```

hsong-rh: 👍

app/policies/request_policy.rb
hsong-rh: @mkanoor Per our discussion online, will refactor this logic after graphql is fixed in insights-api-common.

PR #315, opened 2020-04-02
Query performance optimization

bzwei:
The main purpose is to remove the need of get id list of approver or requester visible resources. All queries are direct. As a result the super long query string with all IDs should not appear in the log.

Prove the requester is now able to see actions. Fixed spec for this query.

### DIFF
--- a/app/controllers/api/v1x0/actions_controller.rb
+++ b/app/controllers/api/v1x0/actions_controller.rb
@@ -5,7 +5,7 @@ module Api
 
       def index
         req = Request.find(params.require(:request_id))
-        authorize req
+        authorize req, :show?
 
         collection(policy_scope(req.actions))
       end
--- a/app/controllers/api/v1x0/requests_controller.rb
+++ b/app/controllers/api/v1x0/requests_controller.rb
@@ -20,7 +20,7 @@ module Api
       def index
         relation = if params[:request_id]
                      req = Request.find(params.require(:request_id))
-                     authorize req
+                     authorize req, :show?
                      req.children
                    else
                      Request
--- a/app/models/request.rb
+++ b/app/models/request.rb
@@ -19,7 +19,6 @@ class Request < ApplicationRecord
 
   scope :decision,       ->(decision)       { where(:decision => decision) }
   scope :state,          ->(state)          { where(:state => state) }
-  scope :owner,          ->(owner)          { where(:owner => owner) }
   scope :requester_name, ->(requester_name) { where(:requester_name => requester_name) }
   scope :group_ref,      ->(group_ref)      { where(:group_ref => group_ref) }
   default_scope { order(:created_at => :desc) }
--- a/app/policies/action_policy.rb
+++ b/app/policies/action_policy.rb
@@ -6,18 +6,9 @@ class ActionPolicy < ApplicationPolicy
 
   class Scope < ApplicationPolicy::Scope
     def resolve
-      raise Exceptions::NotAuthorizedError, "Read access not authorized for #{scope}" unless permission_check('read', scope.model)
-
-      if admin?(scope.model)
-        scope.all
-      elsif approver?(scope.model)
-        scope.where(:id => approver_id_list(scope.model.table_name))
-      elsif requester?(scope.model)
-        scope.where(:id => owner_id_list(scope.table_name))
-      else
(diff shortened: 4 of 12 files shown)


app/policies/mixins/rbac_mixin.rb
```diff
@@ -4,12 +4,15 @@ module RBACMixin
 
     APPROVER_VISIBLE_STATES = [ApprovalStates::NOTIFIED_STATE, ApprovalStates::COMPLETED_STATE].freeze
 
-    def resource_check(verb, id = @record.id, klass = @record.class)
+    def resource_check(verb, record = @record)
       return true unless @user.rbac_enabled?
 
+      klass = record.class
+      return permission_check(verb, klass) unless [Request, Action].include?(klass)
```
hsong-rh, line 11: Don't need `permission_check` here, it's handled by `return false` when no proper role is found

app/policies/mixins/rbac_mixin.rb
```diff
@@ -34,58 +37,28 @@ def requester?(klass = @record.class, verb = 'read')
...
 
-    def owner_request_ids
-      Request.by_owner.pluck(:id).sort
-    end
+    private
 
-    # All child request ids for approver to process
-    def visible_request_ids_for_approver
-      Request.where(:group_ref => @user.group_uuids, :state => APPROVER_VISIBLE_STATES).pluck(:id)
+    def approver_accessible?(record)
```
hsong-rh, line 54: Approver can only read the requests whose states in `APPROVER_VISIBLE_STATES`. This check is missed here.

app/policies/mixins/rbac_mixin.rb
bzwei: It is just a safe check, reminding this method is for Request and Action resource check.

app/controllers/api/v1x0/actions_controller.rb
```diff
@@ -5,7 +5,7 @@ class ActionsController < ApplicationController
 
       def index
         req = Request.find(params.require(:request_id))
-        authorize req
+        authorize req, :show?
```
hsong-rh, line 9: This `authorize` check will be moved to `ActionPolicy` in that user capabilities PR

app/controllers/api/v1x0/requests_controller.rb
```diff
@@ -20,7 +20,7 @@ def show
       def index
         relation = if params[:request_id]
                      req = Request.find(params.require(:request_id))
-                     authorize req
+                     authorize req, :show?
```
hsong-rh, line 23: Same here, move to 'RequestPolicy'

app/policies/action_policy.rb
```diff
@@ -6,18 +6,9 @@ class ActionPolicy < ApplicationPolicy
...
-        scope.where(:id => approver_id_list(scope.model.table_name))
-      elsif requester?(scope.model)
-        scope.where(:id => owner_id_list(scope.table_name))
-      else
-        Rails.logger.error("Error in request resolve: scope does not include admin, group, or user. List of scopes: #{scope.model}")
-        scope.none
-      end
+      # Must through a request
+      raise Exceptions::NotAuthorizedError, "Not authorized to directly access actions" if scope == Action
+      scope
```
hsong-rh, line 11: Do we need this line?

app/controllers/api/v1x0/actions_controller.rb
bzwei: Understand, it is done by your PR.

app/controllers/api/v1x0/requests_controller.rb
bzwei: done in your PR.

app/policies/action_policy.rb
bzwei: Yes. This prevents GraphQL from getting actions from root level. This is why in graphql controller we want to pass `model_class`, not `model_class.all`

app/models/request.rb
```diff
@@ -19,7 +19,7 @@ class Request < ApplicationRecord
 
   scope :decision,       ->(decision)       { where(:decision => decision) }
   scope :state,          ->(state)          { where(:state => state) }
-  scope :owner,          ->(owner)          { where(:owner => owner) }
+  #scope :owner,          ->(owner)          { where(:owner => owner) }
```
hsong-rh, line 22: Let's remove this comment

PR #316, opened 2020-04-02
Add `sort_by` option from common gem to index operations

lindgrenj6:
Same as I did for catalog here: https://github.com/RedHatInsights/catalog-api/pull/647

This allows the user to add `?sort_by=field` to sort the index by any field. 

cc @gmcculloug @bzwei


hsong-rh:
@lindgrenj6 The purpose of this change is for UI?

gmcculloug:
@hsong-rh The UI would likely use it but the purpose is for any API caller, nothing specific for UI here.

lindgrenj6:
@hsong-rh yes, it's included with the common gem. This way the frontend (or any API user) could start asking for the collections ordered by different fields.

PR #317, opened 2020-04-06
Fix to allow graphql to fetch child requests

bzwei:
In GraphQL controller we now pass a scope rather than the model class itself, so that the `parent_id` will be not force to set to `nil` in https://github.com/RedHatInsights/approval-api/blob/stable/app/policies/request_policy.rb#L14


gmcculloug:
cc @zsadeh.  Travis passed but smoke-tests are failing

gmcculloug:
Got approval from @zsadeh to merge with failing smoke-tests which are related to environmental issues.

PR #319, opened 2020-04-07
Fix getting uuid from group_ref

bzwei:
Fix a bug brought in by #310. Before `group_ref` is group's uuid. Now we need to get by `group_ref['uuid']`.

Spec test did not catch the bug because we use rspec's `double` to mock rbac calls. It does not check whether the argument has correct type.

Since we do a group validate at the beginning, we do not need to get the group name from rbac. Instead we can directly use `group_ref['name']`

### DIFF
--- a/app/services/request_create_service.rb
+++ b/app/services/request_create_service.rb
@@ -5,6 +5,7 @@ class RequestCreateService
   attr_accessor :workflows
 
   AUTO_APPROVED_REASON = 'Auto-approved'.freeze
+  SYSTEM_APPROVAL = 'System approval'.freeze
 
   def create(options)
     requester = Insights::API::Common::Request.current.user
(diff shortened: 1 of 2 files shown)


app/services/request_create_service.rb
```diff
@@ -51,15 +51,9 @@ def create_child_requests(request)
...
-    group_name = if group_ref
-                   ContextService.new(leaf_request.context).as_org_admin do
-                     Group.find(group_ref).name
-                   end
-                 else
-                   'System approval'
-                 end
-
-    leaf_request.update!(:workflow_id => workflow_id, :group_ref => group_ref, :group_name => group_name)
+    group_name = group_ref.try(:[], 'name') || 'System approval'
```
gmcculloug, line 54: Recommed making `System approval` a constant in this class and update the spec to reference the constant as well.  https://github.com/RedHatInsights/approval-api/blob/master/spec/services/request_create_service_spec.rb#L182

PR #320, opened 2020-04-08
Added graphql args in resolver

hsong-rh:
Related to PR RedHatInsights/insights-api-common-rails#177

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -5,7 +5,7 @@ require File.join(Bundler::Plugin.index.load_paths("bundler-inject")[0], "bundle
 
 gem 'acts_as_list',        '~> 1.0'
 gem 'faraday',             '>= 0.17.0'
-gem 'insights-api-common', '~> 3.10'
+gem 'insights-api-common', :git => "https://github.com/mkanoor/manageiq-api-common.git", :branch => "support_args"
 gem 'jbuilder',            '~> 2.0'
 gem 'manageiq-loggers',    '~> 0.2'
 gem 'manageiq-messaging',  '~> 0.1'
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -5,7 +5,7 @@ require File.join(Bundler::Plugin.index.load_paths("bundler-inject")[0], "bundle
 
 gem 'acts_as_list',        '~> 1.0'
 gem 'faraday',             '>= 0.17.0'
-gem 'insights-api-common', '~> 3.8'
+gem 'insights-api-common', :git => "https://github.com/mkanoor/manageiq-api-common.git", :branch => "support_args"
```
bzwei, line 8: Do we have to use git? can we wait until next gem releases? There is no blocker issue that we have solve by switching to use a git branch.
We can mark the PR wip until the new gem is released.

app/controllers/api/v1x0/graphql_controller.rb
```diff
@@ -16,7 +16,7 @@ def query
       def overlay
         {
           "^.*$" => {
-            "base_query" => lambda do |model_class, _ctx|
+            "base_query" => lambda do |model_class, graphql_args, _ctx|
               policy_scope(model_class.all)
```
bzwei, line 20: `graphql_args` is not used.
We should switch to use `policy_scope(model_class)`, shouldn't we?

Gemfile
mkanoor: @bzwei That is going to get stored in the user_context, we just wanted to get level set with the manager-common-gem changes and then start using the graphql_args.

app/controllers/api/v1x0/graphql_controller.rb
hsong-rh: yes, the change will be in the next PR.

PR #326, opened 2020-04-15
Set format of group_ref to be uuid

hsong-rh:
https://projects.engineering.redhat.com/browse/SSP-1431

Similar work as https://github.com/RedHatInsights/catalog-api/pull/605


bzwei:
All group_refs need to be in true uuid format in spec tests.

PR #329, opened 2020-04-17
Used GraphQL params to narrow policy scopes

hsong-rh:
https://projects.engineering.redhat.com/browse/SSP-1445

This PR will process graphQL query whose params contain `id`.

### DIFF
--- a/app/controllers/api/v1x0/graphql_controller.rb
+++ b/app/controllers/api/v1x0/graphql_controller.rb
@@ -17,6 +17,7 @@ module Api
         {
           "^.*$" => {
             "base_query" => lambda do |model_class, graphql_args, _ctx|
+              UserContext.current_user_context.graphql_params = graphql_args
               policy_scope(model_class.all)
             end
           }
--- a/app/controllers/api/v1x0/graphql_controller.rb
+++ b/app/controllers/api/v1x0/graphql_controller.rb
@@ -18,7 +18,7 @@ module Api
           "^.*$" => {
             "base_query" => lambda do |model_class, graphql_args, _ctx|
               UserContext.current_user_context.graphql_params = graphql_args
-              policy_scope(model_class.all)
+              policy_scope(model_class)
             end
           }
         }
--- a/app/models/request.rb
+++ b/app/models/request.rb
@@ -11,8 +11,9 @@ class Request < ApplicationRecord
   has_many :actions, -> { order(:id => :asc) }, :dependent => :destroy, :inverse_of => :request
   has_many :random_access_keys, :dependent => :destroy, :inverse_of => :request
 
-  belongs_to :parent,   :foreign_key => :parent_id, :class_name => 'Request', :inverse_of => :children
-  has_many   :children, :foreign_key => :parent_id, :class_name => 'Request', :inverse_of => :parent, :dependent => :destroy
+  belongs_to :parent,   :foreign_key => :parent_id, :class_name => 'Request', :inverse_of => :requests
+  has_many   :requests, :foreign_key => :parent_id, :class_name => 'Request', :inverse_of => :parent, :dependent => :destroy
+  alias_method :children, :requests
 
   validates :name,     :presence  => true
   validates :state,    :inclusion => { :in => STATES }
--- a/app/models/request.rb
+++ b/app/models/request.rb
@@ -22,6 +22,7 @@ class Request < ApplicationRecord
   scope :state,          ->(state)          { where(:state => state) }
   scope :requester_name, ->(requester_name) { where(:requester_name => requester_name) }
   scope :group_ref,      ->(group_ref)      { where(:group_ref => group_ref) }
+  scope :root_requests,  ->                 { where(:parent_id => nil) }
   default_scope { order(:created_at => :desc) }
 
   delegate :content, :to => :request_context
--- a/app/models/request.rb
+++ b/app/models/request.rb
@@ -13,7 +13,6 @@ class Request < ApplicationRecord
 
   belongs_to :parent,   :foreign_key => :parent_id, :class_name => 'Request', :inverse_of => :requests
   has_many   :requests, :foreign_key => :parent_id, :class_name => 'Request', :inverse_of => :parent, :dependent => :destroy
-  alias_method :children, :requests
 
   validates :name,     :presence  => true
   validates :state,    :inclusion => { :in => STATES }
@@ -31,13 +30,13 @@ class Request < ApplicationRecord
   after_initialize :set_defaults
 
   def invalidate_number_of_children
-    update(:number_of_children => children.size)
+    update(:number_of_children => requests.size)
   end
 
   def invalidate_number_of_finished_children
     return if number_of_children.zero?
 
-    update(:number_of_finished_children => children.count(&:finished?))
+    update(:number_of_finished_children => requests.count(&:finished?))
   end
 
   def create_child
--- a/app/policies/action_policy.rb
+++ b/app/policies/action_policy.rb
@@ -1,6 +1,8 @@
 class ActionPolicy < ApplicationPolicy
   class Scope < ApplicationPolicy::Scope
     def resolve
+      return graphql_id_query if graphql_query_by_id?
+
       if user.params[:request_id]
         req = Request.find(user.params[:request_id])
         raise Exceptions::NotAuthorizedError, "Read access not authorized for request #{req.id}" unless resource_check('read', req)
--- a/app/policies/action_policy.rb
+++ b/app/policies/action_policy.rb
@@ -1,14 +1,12 @@
 class ActionPolicy < ApplicationPolicy
   class Scope < ApplicationPolicy::Scope
-    def resolve
-      return graphql_id_query if graphql_query_by_id?
-
+    def resolve_scope
       if user.params[:request_id]
         req = Request.find(user.params[:request_id])
         raise Exceptions::NotAuthorizedError, "Read access not authorized for request #{req.id}" unless resource_check('read', req)
 
         req.actions
-      else # Both GraphQL and direct access are not allowed 
+      else # Both GraphQL and direct access are not allowed
         raise Exceptions::NotAuthorizedError, "Not authorized to directly access actions"
       end
     end
@@ -28,11 +26,9 @@ class ActionPolicy < ApplicationPolicy
     operation = user.params.require(:operation)
     uuid = user.request.headers['x-rh-random-access-key']
 
-    valid_operation =
-      admin?(record, 'create') && Action::ADMIN_OPERATIONS.include?(operation) ||
+    admin?(record, 'create') && Action::ADMIN_OPERATIONS.include?(operation) ||
       approver?(record, 'create') && Action::APPROVER_OPERATIONS.include?(operation) ||
       requester?(record, 'create') && Action::REQUESTER_OPERATIONS.include?(operation) ||
-
       uuid.present? && Request.find(user.params[:request_id]).try(:random_access_keys).any? { |key| key.access_key == uuid }
   end
 end
--- a/app/policies/application_policy.rb
+++ b/app/policies/application_policy.rb
@@ -59,5 +59,34 @@ class ApplicationPolicy
     def resolve
       scope.all
     end
+
+    def graphql_id_query
+      id = user.graphql_params.id
+
+      item = scope.find(id)
+      raise Exceptions::NotAuthorizedError, "Read access not authorized for request #{item.id}" unless resource_check('read', item)
+
+      scope.where(:id => id)
+    end
+
+    def graphql_filter_query
+      scope.where(user.graphql_params.filter)
+    end
+
+    def graphql_collection_query
+      graphql_query_by_filter? ? graphql_filter_query : scope.all
+    end
+
+    def graphql_query_by_id?
+      graphql_query? && user.graphql_params.id.present?
+    end
+
+    def graphql_query_by_filter?
+      graphql_query? && user.graphql_params.filter.present?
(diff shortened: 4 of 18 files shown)


app/policies/request_policy.rb
```diff
@@ -6,24 +6,23 @@ class Scope < ApplicationPolicy::Scope
 
     def resolve
       return scope.all unless user.rbac_enabled?
+      return graphql_id_query if graphql_query_by_id?
 
       if user.params[:request_id]
         req = Request.find(user.params[:request_id])
         raise Exceptions::NotAuthorizedError, "Read access not authorized for request #{req.id}" unless resource_check('read', req)
         req.children
```
mkanoor, line 14: @hsong-rh Should we change this to req.requests instead of children.

app/policies/action_policy.rb
```diff
@@ -1,6 +1,8 @@
 class ActionPolicy < ApplicationPolicy
   class Scope < ApplicationPolicy::Scope
     def resolve
+      return graphql_id_query if graphql_query_by_id?
+
       if user.params[:request_id]
```
mkanoor, line 6: @hsong-rh Why not use
```
if graphql_query_by_id?
 graphql_id_query
else if user.params[:request_id]
....
```

Otherwise it looks like an afterthought in the design process

spec/controllers/v1.0/graphql_controller_spec.rb
```diff
@@ -45,25 +52,35 @@
...
+    let(:user) { instance_double(UserContext, :access => access, :rbac_enabled? => true, :params => params, :graphql_params => graphql_params) }
+    let(:graphql_params) { double("graphql_params", :id => "#{parent_request.id}", :filter => nil) }
 
     context 'requests with admin role' do
       let(:headers) { headers_with_admin }
       before { admin_access }
 
       it 'return requests' do
-        post "#{api_version}/graphql", :headers => headers, :params => graphql_requests_query
+        post "#{api_version}/graphql", :headers => headers, :params => graphql_id_query
```
mkanoor, line 70: @hsong-rh Do you have a case where where one ID is visible but another ID is not accessible?

app/policies/action_policy.rb
bzwei: @mkanoor I believe `return graphql_id_query if graphql_query_by_id?` is a recommended style rubocop prefers.

app/models/request.rb
```diff
@@ -30,13 +30,13 @@ class Request < ApplicationRecord
   after_initialize :set_defaults
 
   def invalidate_number_of_children
-    update(:number_of_children => children.size)
+    update(:number_of_children => requests.size)
```
bzwei, line 33: why not create an alias method `children`?

app/policies/application_policy.rb
```diff
@@ -59,5 +59,34 @@ def initialize(user, scope)
...
+      raise Exceptions::NotAuthorizedError, "Read access not authorized for request #{item.id}" unless resource_check('read', item)
+
+      scope.where(:id => id)
+    end
+
+    def graphql_filter_query
+      scope.where(user.graphql_params.filter)
+    end
+
+    def graphql_collection_query
```
bzwei, line 76: you only need
```
  def graphql_collection_query
    scope.where(user.graphql_params&.filter)
  end
```

Methods `graphql_query_by_filter?` and `graphql_filter_query` can go away.

app/models/request.rb
mkanoor: @bzwei Because it doesn't reflect reality
The URL reads /requests/{id}/requests
https://github.com/RedHatInsights/approval-api/blob/bf3bd885938f791012835d660e166e5057e96119/public/doc/openapi-3-v1.2.json#L331
But internally you use children, when the graphql parser generates the schema from the openapi.json it doesn't know what children is. So this is needed for consistency

bzwei: that's exactly what I meant. We can use children as an alias method.

app/policies/application_policy.rb
hsong-rh: 👍

spec/controllers/v1.0/graphql_controller_spec.rb
hsong-rh: Just added a case for regular user. Now one is visible, the other is not

app/models/request.rb
hsong-rh: @bzwei Remove extra to keep codes simple

app/policies/request_policy.rb
```diff
@@ -6,24 +6,23 @@ class Scope < ApplicationPolicy::Scope
 
     def resolve
       return scope.all unless user.rbac_enabled?
```
bzwei, line 8: why do we need this line here while all other policies do not have it?

app/policies/template_policy.rb
```diff
@@ -1,7 +1,10 @@
 class TemplatePolicy < ApplicationPolicy
   class Scope < ApplicationPolicy::Scope
     def resolve
```
bzwei, line 3: This implementation is the same as for workflow policy. We can make it a default implementation in `ApplicationPolicy::Scope`.

app/policies/template_policy.rb
```diff
@@ -1,7 +1,10 @@
...
-      permission_check('read', scope) ? scope.all : (raise Exceptions::NotAuthorizedError, "Read access not authorized for #{scope}")
+      return graphql_id_query if graphql_query_by_id?
```
bzwei, line 4: This line repeats in every scope's `resolve` method. We can move it in to its base implementation. In base's `resolve` method we can call `resolve_graphql_scope` and `resolve_native_scope`. And you only need to override `resolve_native_scope` for request and action policies.

bzwei: Actually you even don't need `resolve_graphql_scope`. The base scope's resolve method can be
```
  def resolve
    return graphql_id_query if graphql_query_by_id?
   resolve_collection_scope
  end

  def resolve_collection_scope
     raise Exceptions::NotAuthorizedError, "Read access not authorized for #{scope}" unless permission_check('read', scope)

    graphql_query? ? graphql_collection_query : scope.all
  end
```
Seems `resolve_collection_scope` is a better name.

app/policies/request_policy.rb
```diff
@@ -4,26 +4,21 @@ class Scope < ApplicationPolicy::Scope
...
-        req.children
-      # GraphQL comes in with model_class.all, whose scope is respond to model call
-      # Regular root requests call comes in with model class itself
+        req.requests
       else
         case Insights::API::Common::Request.current.headers[Insights::API::Common::Request::PERSONA_KEY]
         when PERSONA_ADMIN
           raise Exceptions::NotAuthorizedError, "No permission to access the complete list of requests" unless admin?(scope)
-          scope.respond_to?(:model) ? scope : scope.where(:parent_id => nil)
+          scope.where(:parent_id => nil)
```
mkanoor, line 16: @hsong-rh Does it make sense to add a scope in the app/model/requests for top_level_requests so we dont have to have parent_id => null logic in multiple places.

Similar to

https://github.com/RedHatInsights/approval-api/blob/5fe66adc87f6df620bbdec90ba750beb3dd8025e/app/models/concerns/owner_field.rb#L8

app/policies/application_policy.rb
```diff
@@ -57,7 +57,32 @@ def initialize(user, scope)
     end
 
     def resolve
+      return scope.all unless user.rbac_enabled?
+      return  graphql_id_query if graphql_query_by_id?
```
mkanoor, line 61: @hsong-rh 
can this be a ternary
```
graphql_query_by_id? ? graphql_id_query : resolve_scope
```

app/policies/application_policy.rb
```diff
@@ -57,7 +57,32 @@ def initialize(user, scope)
     end
 
     def resolve
+      return scope.all unless user.rbac_enabled?
```
bzwei, line 60:
```
  def resolve
    return graphql_id_query if graphql_query_by_id

    return scope.all unless user.rbac_enabled?

    resolve_scope
  end
```

app/policies/application_policy.rb
```diff
@@ -57,7 +57,32 @@ def initialize(user, scope)
...
+
+      graphql_query_by_id? ? graphql_id_query : resolve_scope
+    end
+
+    def resolve_scope
+      raise Exceptions::NotAuthorizedError, "Read access not authorized for #{scope}" unless permission_check('read', scope)
```
bzwei, line 66: should this check part of `resolve` method?

app/policies/application_policy.rb
```diff
@@ -57,7 +57,32 @@ def initialize(user, scope)
...
       scope.all
     end
+
+    def graphql_id_query
+      id = user.graphql_params.id
+
+      item = scope.find(id)
+      raise Exceptions::NotAuthorizedError, "Read access not authorized for request #{item.id}" unless resource_check('read', item)
+
+      scope.where(:id => id)
```
bzwei, line 77: no sure why we need to add `where(:id => id)`? Wont' graphql implementation add it just like graphql_params.filters are handled?

app/policies/application_policy.rb
```diff
@@ -57,7 +57,32 @@ def initialize(user, scope)
...
+    end
+
+    def graphql_query_by_id?
+      graphql_query? && user.graphql_params.id.present?
```
bzwei, line 81:
```
  user.graphql_params&.id # or !!user.graphql_params&.id
```
seems sufficient. We don't need `graphql_query?`

app/policies/request_policy.rb
```diff
@@ -4,26 +4,24 @@ class Scope < ApplicationPolicy::Scope
...
           raise Exceptions::NotAuthorizedError, "No permission to access the complete list of requests" unless admin?(scope)
-          scope.respond_to?(:model) ? scope : scope.where(:parent_id => nil)
+
+          scope.root_requests_only
         when PERSONA_APPROVER
           raise Exceptions::NotAuthorizedError, "No permission to access requests assigned to approvers" unless approver?(scope)
-          scope.respond_to?(:model) ? (raise Exceptions::NotAuthorizedError, "Approver not allowed to access requests via GraphQL") :
-                                      approver_visible_requests(scope)
+
+          approver_visible_requests(scope).child_requests_only
```
bzwei, line 22: should not and cannot have `.child_requests_only` because qualified requests may include root requests if they don't have children.

app/policies/application_policy.rb
hsong-rh: `where(:id => id)` will help to populate the return results with their associated resources, ex. every child requests and their actions

app/policies/application_policy.rb
hsong-rh: No, `request` and `action` don't do this check

app/policies/application_policy.rb
hsong-rh: `graphql_id_query`  depends on rbac check. it will fail if rbac is disabled.

app/policies/request_policy.rb
hsong-rh: 👍

app/policies/application_policy.rb
bzwei: but you cannot return `scope.all` if the query is for one id

app/policies/application_policy.rb
bzwei: GraphQL should still work if we don't have rbac enabled. So it lead me to conclude `where(:id=>id)` is not needed.

app/policies/application_policy.rb
hsong-rh: `resolve` is always called by `index` to return a list. Even for `graphql_query_by_id`, it returns an association with single instance.

app/policies/application_policy.rb
```diff
@@ -57,7 +57,28 @@ def initialize(user, scope)
...
+      id = user.graphql_params.id
+
+      item = scope.find(id)
+      raise Exceptions::NotAuthorizedError, "Read access not authorized for request #{item.id}" unless resource_check('read', item)
+
+      scope.where(:id => id)
+    end
+
+    def graphql_query_by_id?
+      user.graphql_params&.id && user.graphql_params.id.present?
```
bzwei, line 81: no need the second part.

app/policies/application_policy.rb
hsong-rh: yes, GraphQL should work, but here we only send one record to it, rather than a collection, feel we should have better performance

app/policies/request_policy.rb
```diff
@@ -4,26 +4,24 @@ class Scope < ApplicationPolicy::Scope
...
-      # Regular root requests call comes in with model class itself
+
+        req.requests
       else
         case Insights::API::Common::Request.current.headers[Insights::API::Common::Request::PERSONA_KEY]
         when PERSONA_ADMIN
           raise Exceptions::NotAuthorizedError, "No permission to access the complete list of requests" unless admin?(scope)
-          scope.respond_to?(:model) ? scope : scope.where(:parent_id => nil)
+
+          scope.root_requests_only
```
mkanoor, line 18: @hsong-rh can this be just root_requests we dont need the only.

app/controllers/api/v1x0/graphql_controller.rb
```diff
@@ -17,6 +17,7 @@ def overlay
         {
           "^.*$" => {
             "base_query" => lambda do |model_class, graphql_args, _ctx|
+              UserContext.current_user_context.graphql_params = graphql_args
               policy_scope(model_class.all)
```
bzwei, line 21: shall we remove `.all`?

PR #330, opened 2020-04-20
Used class method to cache PolicyClass name

hsong-rh:
Same as https://github.com/RedHatInsights/catalog-api/pull/690

### DIFF
--- a/app/models/concerns/metadata.rb
+++ b/app/models/concerns/metadata.rb
@@ -18,6 +18,8 @@ module Metadata
   private
 
   def user_capabilities
-    user_context.nil? ? { } : self.class.policy_class.new(user_context, self).user_capabilities
+    return nil if user_context.nil?
+
+    self.class.policy_class.new(user_context, self).user_capabilities
   end
 end
--- a/app/models/concerns/metadata.rb
+++ b/app/models/concerns/metadata.rb
@@ -3,17 +3,21 @@ module Metadata
 
   included do
     attribute :metadata, ActiveRecord::Type::Json.new
+  end
 
-    def metadata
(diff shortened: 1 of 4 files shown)


app/models/concerns/metadata.rb
```diff
@@ -3,17 +3,21 @@ module Metadata
...
+    { :user_capabilities => user_capabilities }
+  end
 
-    def policy_class
-      "#{self.class}Policy".constantize
-    end
+  private
+
+  def user_capabilities
+    user_context.nil? ? { } : self.class.policy_class.new(user_context, self).user_capabilities
```
gmcculloug, line 21: While this is similar to the change from RedHatInsights/catalog-api#690 there is an important different in `user_capabilities` that I want to ensure makes sense.

I catalog it returns `nil` if `user_context` is nil
```ruby
def user_capabilities
  return nil if user_context.nil?

  "#{self.class}Policy".constantize.new(user_context, self).user_capabilities
  self.class.policy_class.new(user_context, self).user_capabilities
end
```
But here it is returning `{}`.

@mkanoor @hsong-rh Does this different make sense or should we make them the same?  And if this method changes to return `nil` what is the impact to the approval code base?

mkanoor: @gmcculloug Make sense to have them be the same. The nil versus {} comes in only in dev mode when you are running this from the rails console. Otherwise we could always be running with a UserContext and will never see a nil or {}. The UserContext and the user_capabilities code are almost similar in catalog and approval services

hsong-rh: @gmcculloug Changed to return `nil`, make it same as catalog

PR #333, opened 2020-04-29
Extend approver group validation to admin group

hsong-rh:
Allow admin groups to be added to approval process

https://projects.engineering.redhat.com/browse/SSP-1435

### DIFF
--- a/app/models/group.rb
+++ b/app/models/group.rb
@@ -31,8 +31,41 @@ class Group
     @roles ||= Group.find(uuid).roles
   end
 
-  def has_role?(role_name)
-    roles.any? { |role| role.name == role_name }
+  def can_approve?
+    context = Insights::API::Common::Request.current.to_h.transform_keys(&:to_s)
+    @can_approve ||= ContextService.new(context).as_org_admin do
+      (action_create_scopes & ['admin', 'group']).any?
+    end
+  end
+
+  private
+
+  def action_create_scopes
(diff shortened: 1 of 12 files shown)


app/models/group.rb
```diff
@@ -31,8 +31,41 @@ def roles
     @roles ||= Group.find(uuid).roles
   end
 
-  def has_role?(role_name)
-    roles.any? { |role| role.name == role_name }
+  def can_approve?
```
bzwei, line 34: cache the result
```
def can_approve?
  return @can_approve unless @can_approve.nil?
  @can_approve = ...
end
```

app/models/group.rb
```diff
@@ -31,8 +31,41 @@ def roles
...
+    context = Insights::API::Common::Request.current.to_h.transform_keys(&:to_s)
+    ContextService.new(context).as_org_admin do
+      (scopes & ['admin', 'group']).any?
+    end
+  end
+
+  private
+
+  def scopes
```
bzwei, line 43: rename `action_create_scopes` because you are collecting scopes for specific permissions only

app/models/group.rb
eclarizio: Should just be able to use `@can_approval ||=`

bzwei:
@hsong-rh It looks good now. For the testing, can we add one to test the `can_approve?` itself since the existing ones only simply stub the method?

hsong-rh:
@bzwei A simple test is added

PR #339, opened 2020-05-05
Drop unused AccessControlEntry table

hsong-rh:


db/schema.rb
```diff
@@ -78,6 +65,7 @@
     t.string "owner"
     t.bigint "parent_id"
     t.bigint "request_context_id"
+    t.string "random_access_key"
```
bzwei, line 68: should not have this change

db/schema.rb
```diff
@@ -86,6 +74,7 @@
     t.datetime "finished_at"
     t.index ["group_ref"], name: "index_requests_on_group_ref"
     t.index ["parent_id"], name: "index_requests_on_parent_id"
+    t.index ["random_access_key"], name: "index_requests_on_random_access_key"
```
bzwei, line 77: should not have it

gmcculloug:
@eclarizio Please review

db/schema.rb
hsong-rh: Removed these auto generated changes

PR #341, opened 2020-05-06
Refactored model related spec tests

hsong-rh:


spec/models/request_spec.rb
```diff
@@ -113,6 +113,18 @@
     end
   end
 
+  context '#group' do
```
bzwei, line 116: describe

spec/models/request_spec.rb
```diff
@@ -113,6 +113,18 @@
...
+    let(:group) { instance_double(Group, :name => 'g1') }
+    before { allow(Group).to receive(:find).and_return(group) }
+
+    it 'returns group' do
+      expect(Group).to receive(:find).once
+      expect(subject.group).to eq(group)
+
+      subject.group
```
bzwei, line 124: not needed, right?

PR #343, opened 2020-05-07
Drop unused v1.0 and v1.1

hsong-rh:
Early versions 1.0 and 1.1 are unusable due to db schema changes, such as separating `random_access_key` into its own table; different format of `group_ref` for `workflow`, etc.

So clean them up here.


gmcculloug:
@hsong-rh Can you give a more details in the description about what changed that led to this?  

@eclarizio @mkanoor Curious your thoughts on dropping these minor versions.

hsong-rh:
@eclarizio Good catch, I miss them and will change.

mkanoor:
@hsong-rh Do we need to regenerate the client
It still says v1.0.0
https://github.com/RedHatInsights/approval-api-client-ruby/blob/5bdb4c7fecc44693ada76f93b8cd60d1a7c8f241/approval-api-client-ruby.gemspec#L8

mkanoor:
@hsong-rh Should the version string in the openapi.json be adjusted to say 1.2.0
https://github.com/RedHatInsights/approval-api-client-ruby/blob/5bdb4c7fecc44693ada76f93b8cd60d1a7c8f241/openapi.json#L6

mkanoor:
@hsong-rh In QA approval logs I found
https://qa.cloud.redhat.com/api/approval/v1.0/templates

Will this automatically get routed to v1.2/templates once we remove all the v1.0 code.

hsong-rh:
@mkanoor yes, we will regenerate the client after Bill updates the openapi. All v1.0/v1.1 calls will automatically route to v1.2

hsong-rh:
@eclarizio I changed to redirect v1.0 and v1.1 calls to v1.2

PR #346, opened 2020-05-08
Add Github .gitattributes file

gmcculloug:
The [library that Github](https://help.github.com/en/github/creating-cloning-and-archiving-repositories/about-repository-languages) uses to detect a repo language is detect this repo as being mostly CSS, when in reality it would be more accurate to display as a Ruby repo.

image

This `.gitattributes` file will cause the library to ignore CSS files and should promote the repo and mainly a ruby based repo.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+* linguist-vendored
+*.css linguist-vendored=false


gmcculloug:
cc @epwinchell

PR #348, opened 2020-05-11
Added older versions of openapi.json files back

hsong-rh:
In order to support redirecting to older version (`v1.0` and `v1.1`), their openapi json files are still pointed in api common gem and needed to add back.

### DIFF
--- a/public/doc/openapi-3-v1.0.json
+++ b/public/doc/openapi-3-v1.0.json
@@ -1,0 +1,1540 @@
+{
+    "openapi": "3.0.0",
+    "info": {
+        "title": "Insights Service Approval APIs",
+        "description": "APIs to query approval service",
+        "version": "1.0.0",
+        "contact": {
+            "email": "support@redhat.com"
+        },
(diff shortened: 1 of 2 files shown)


bzwei:
Do we still use v1.0 and v1.1 in any place? If not, we can remove redirect also.
cc @lgalis @zsadeh @eclarizio

zsadeh:
@bzwei I'm only using latest API version.

gmcculloug:
The Catalog UI had a call to v1.0 but that was corrected this morning: https://github.com/RedHatInsights/catalog-ui/pull/572

@eclarizio @mkanoor Any thoughts on if this is needed?

eclarizio:
@gmcculloug We just use the approval-api-client-ruby, so assuming that is up to date to automatically use the latest version, then I don't think we have anywhere we are actively calling v1.0.

hsong-rh:
Based on the feedback, no need to keep redirect logic. PR closed.

PR #349, opened 2020-05-11
Make sequence unique

bzwei:
Remove dependency on acts_as_list. Implement a simple logic for sequence.

https://projects.engineering.redhat.com/browse/SSP-1471

### DIFF
--- a/app/models/workflow.rb
+++ b/app/models/workflow.rb
@@ -2,7 +2,6 @@ class Workflow < ApplicationRecord
   include Metadata
   acts_as_tenant(:tenant)
 
-  acts_as_list :scope => [:tenant_id], :column => 'sequence'
   default_scope { order(:sequence => :asc) }
 
   belongs_to :template
@@ -10,6 +9,10 @@ class Workflow < ApplicationRecord
   has_many :tag_links, :dependent => :destroy, :inverse_of => :workflow
 
   validates :name, :presence => true, :uniqueness => {:scope => :tenant}
+  validates :sequence, :uniqueness => { scope: :tenant_id }
+
+  before_validation :new_sequence, :on => :create
+  before_validation :move_rest_higher, :on => :update
 
   def external_processing?
     template&.process_setting.present?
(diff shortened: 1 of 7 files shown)


gmcculloug:
@bzwei What happens when you delete one?  Will it create gaps that never get filled in?  Since the number is visible that seems like an issue.

bzwei:
@gmcculloug Gap is allowed even with `acts_as_list` that we used before. What guaranteed is the sequence order. 
Not only the delete method creates a gap, the adjustment of sequence also generates a gap. I think it is OK. We could try to avoid gaps by conducting more `update_all` calls with a tradeoff of performance.

app/models/workflow.rb
```diff
@@ -18,4 +21,24 @@ def external_processing?
...
+
+    self.sequence = self.class.last&.sequence.to_i + 1
+  end
+
+  # move all related sequence number one higher if the desired number is in use
+  def move_rest_higher
+    return unless sequence && self.class.where(:sequence => sequence).exists
+
+    self.class.where(table[:sequence].gteq(sequence)).update_all("sequence = (-sequence - 1)")
+    self.class.where(table[:sequence].lt(0)).update_all("sequence = (-sequence)")
```
hsong-rh, line 42: should these two steps are `locked` to prevent concurrent issue?

bzwei: try to avoid locking by using transaction and retry.

hsong-rh: I saw it in workflow `update` only, but no in `create`, any reason?

hsong-rh: I saw it in workflow `update` only, but no in `create`, any reason?

app/models/workflow.rb
```diff
@@ -18,4 +21,24 @@ def external_processing?
...
   end
+
+  private
+
+  def table
+    self.class.arel_table
+  end
+
+  def new_sequence
+    return move_rest_higher if sequence
```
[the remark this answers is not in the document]
bzwei: @mkanoor sequence will be auto generated if it is not given in the attribute list when a new workflow is created. However the sequence will be honored if it is given and thus need to push other sequences higher if there is a conflict.

app/models/workflow.rb
bzwei: create by default will always place the new workflow to the end of the sequencing, unless the number is given by caller and this method does get called.

BTW, transaction is not needed in our service classes. See the document https://guides.rubyonrails.org/active_record_callbacks.html#halting-execution. I am going to remove them.

hsong-rh: 👍

PR #352, opened 2020-05-14
Cleanup unused redirecting files to older versions

hsong-rh:
Removed all unneeded files for routes redirecting to older versions


spec/controllers/v1.2/actions_controller_spec.rb
```diff
@@ -1,4 +1,4 @@
-RSpec.describe Api::V1x2::ActionsController, :type => :request do
+RSpec.describe Api::V1x2::ActionsController, :type => [:request, :v1x2] do
```
bzwei, line 1: we can continue to use configurable version other than hard code `:v1x2` here

hsong-rh: This is not hardcoded, eventually it points to:
https://github.com/RedHatInsights/approval-api/blob/cb4f7a7d80ae11c09f582d99c13e5700c6a37538/spec/support/v1x2_helper.rb#L3

PR #353, opened 2020-05-14
Adjust sequences only when current value changes

bzwei:
We should not adjust sequences if current value does not change. Otherwise each save causes the numbers moving higher.

https://projects.engineering.redhat.com/browse/SSP-1536

The above ticket reported the problem occurs at GET. It is because each get does validation on  groups and save with updated group names.

It also resolves a warning on deprecated call.


bzwei:
cc @lgalis

PR #354, opened 2020-05-15
Added missing spec test cases to improve coverage

hsong-rh:


spec/controllers/v1.2/stageaction_controller_spec.rb
```diff
@@ -1,6 +1,72 @@
...
+RSpec.describe Api::V1x2::StageactionController, :type => [:v1x2, :request, :controller] do
+  let(:key) { create(:random_access_key, :access_key => 'unique-123', :approver_name => 'Joe Smith') }
+  let!(:approval_request) { create(:request, :with_tenant, :with_context, :random_access_keys => [key]) }
+
+  describe 'GET /stageaction/:id' do
+    it 'returns 200' do
+      get "#{api_version}/stageaction/#{key.access_key}"
+
+      expect(controller.instance_variable_get(:@request)).to eq(approval_request)
+      expect(controller.instance_variable_get(:@approver)).to eq('Joe Smith')
```
eclarizio, line 10: These expectations don't really match what the `it` block claims this is doing. Looking at the actual code, it appears the `show` is empty. Should we actually be expecting a 200 here or is this just kind of a no-op?

spec/lib/kie/service_spec.rb
```diff
@@ -0,0 +1,14 @@
...
+  let(:options) { {} }
+
+  describe '.call' do
+    it 'raises a KieClient::ApiError' do
+      expect do
+        described_class.call(KieClient::ProcessInstancesBPMApi, options) do |_api|
+          raise kie_ex
+        end
+      end.to raise_exception(Exceptions::KieError)
+    end
```
eclarizio, line 12: I think we might want another spec here testing this line https://github.com/RedHatInsights/approval-api/blob/master/lib/kie/service.rb#L24.

We should be able to do something like this, I think?
```
it "sets the authorization headers" do
  auth_headers = described_class.call(KieClient::ProcessInstancesBPMApi, options) do |api|
    api.api_client.default_headers["Authorization"]
  end

  expect(auth_headers).to eq("<Whatever the basic auth token is>")
end
```

I'm just thinking that if that line changes, the current spec doesn't really test anything about it and other than setting up the `KieClient` configuration, that seems to be an important step that we're not covering.

spec/services/action_create_service_spec.rb
```diff
@@ -145,6 +145,21 @@
...
+
+    it 'forbids skip operation from the stage other than pending' do
+      child1.update(:state => Request::NOTIFIED_STATE)
+      expect { svc1.create('operation' => Action::SKIP_OPERATION, 'processed_by' => 'man') }.to raise_error(Exceptions::InvalidStateTransitionError)
+    end
+
+    it 'forbids notify operation from the stage other than started' do
+      child1.update(:state => Request::NOTIFIED_STATE)
+      expect { svc1.create('operation' => Action::NOTIFY_OPERATION, 'processed_by' => 'man') }.to raise_error(Exceptions::InvalidStateTransitionError)
+    end
```
eclarizio, line 161: A suggestion here (though not necessary), would be to use a context, and you could even use shared examples to reduce some duplication:
```
context "when the child is in a notified state" do
  before do
    child1.update(:state => Request::NOTIFIED_STATE)
  end

  shared_examples_for "forbidden_operation" do |operation|
    it "forbids #{operation}" do
      expect { svc1.create('operation' => operation, 'processed_by' => 'man') }.to raise_error(Exceptions::InvalidStateTransitionError)
    end
  end

  [Action::START_OPERATION, Action::SKIP_OPERATION, Action::NOTIFY_OPERATION].each do |operation|
    it_behaves_like "forbidden_operation", operation
  end
end
```

spec/services/jbpm_process_service_spec.rb
```diff
@@ -64,20 +63,39 @@
...
 
-  context 'when kie service raise exception' do
-    before do
-      allow(jbpm).to receive(:start_process).and_raise(kie_ex)
-      allow(subject).to receive(:enhance_groups)
+  describe 'kie service raise exception' do
+    context 'when start_process' do
+      before do
+        allow(jbpm).to receive(:start_process).and_raise(kie_ex)
+        allow(subject).to receive(:enhance_groups)
```
eclarizio, line 70: I know you just moved this so I don't think we need to change it here, but stubbing out a private method is something I'd like to shy away from. It tells me that either this class is trying to do too much and it's complicated to test, or that there's some functionality that isn't really that important for this test and it could be broken out somewhere else. Just something to keep in mind.

spec/services/jbpm_process_service_spec.rb
```diff
@@ -64,20 +63,39 @@
     end
   end
 
-  context 'when kie service raise exception' do
-    before do
-      allow(jbpm).to receive(:start_process).and_raise(kie_ex)
-      allow(subject).to receive(:enhance_groups)
+  describe 'kie service raise exception' do
+    context 'when start_process' do
```
bzwei, line 67: describe '#start' do

spec/services/jbpm_process_service_spec.rb
```diff
@@ -64,20 +63,39 @@
...
+
+        expect(request.state).to eq(Request::FAILED_STATE)
+        expect(request.decision).to eq(Request::ERROR_STATUS)
+        expect(request.reason).to eq(kie_ex.message)
+      end
     end
 
-    it 'should post an error action' do
-      expect { subject.start }.to raise_exception(Exceptions::KieError)
+    context 'when signal_process_instance' do
```
bzwei, line 84: describe '#signal' do

spec/services/jbpm_process_service_spec.rb
hsong-rh: 👍

spec/controllers/v1.2/stageaction_controller_spec.rb
hsong-rh: Don't know why 500 is always returned here, but in my env it returns 200.

PR #355, opened 2020-05-15
Avoid gaps when adjust sequences

bzwei:
A followup of #353 and #349 

Optimize the algorithm to avoid gaps when update the sequence.

User provided sequence may not be honored. It will be adjusted to the smallest available sequence if desired number is too large.

### DIFF
--- a/app/models/workflow.rb
+++ b/app/models/workflow.rb
@@ -12,7 +12,8 @@ class Workflow < ApplicationRecord
   validates :sequence, :uniqueness => { scope: :tenant_id }
 
   before_validation :new_sequence, :on => :create
-  before_validation :move_rest_higher, :on => :update
+  before_validation :adjust_sequences, :on => :update
+  before_destroy    :sequence_lower
 
   def external_processing?
     template&.process_setting.present?
@@ -29,16 +30,54 @@ class Workflow < ApplicationRecord
   end
 
   def new_sequence
-    return move_rest_higher if sequence
+    throw :abort if sequence && sequence <= 0
 
-    self.sequence = self.class.last&.sequence.to_i + 1
+    largest = last_sequence
+    if sequence && sequence < last_sequence
+      sequence_higher(sequence)
+    else
+      self.sequence = largest + 1 # auto_assignment if sequence is nil or too large
+    end
   end
 
-  # move all related sequence number one higher if the desired number is in use
-  def move_rest_higher
-    return unless sequence && sequence_changed? && self.class.exists?(:sequence => sequence)
+  # no gap between sequences
+  def adjust_sequences
+    return unless sequence_changed?
 
-    self.class.where(table[:sequence].gteq(sequence)).update_all("sequence = (-sequence - 1)")
-    self.class.where(table[:sequence].lt(0)).update_all("sequence = (-sequence)")
+    throw :abort if sequence && sequence <= 0
+
+    largest = last_sequence
+    self.sequence = largest if sequence.nil? || sequence > largest
+
+    if sequence > sequence_was
+      sequence_lower(sequence_was, sequence)
+    else
+      sequence_higher(sequence, sequence_was)
+    end
+  end
+
+  # sequences increment between [sequence sequence_was sequence)
+  def sequence_higher(startp, endp = nil)
+    query = self.class.where(table[:sequence].gteq(startp))
+    query = query.where(table[:sequence].lteq(endp)) if endp
+    query.update_all("sequence = (-sequence - 1)")
+
+    query = self.class.where(table[:sequence].lt(0))
+    query = query.where.not(:sequence => -endp - 1) if endp # endp to be updated by rails
+    query.update_all("sequence = (-sequence)")
+  end
+
+  # sequences decrement between [startp endp]
+  def sequence_lower(startp = sequence, endp = nil)
+    query = self.class.where(table[:sequence].gteq(startp))
+    query = query.where(table[:sequence].lteq(endp)) if endp
+    query.update_all("sequence = (-sequence + 1)")
+
+    # startp to be updated by rails
+    query = self.class.where(table[:sequence].lt(0)).where.not(:sequence => -startp + 1).update_all("sequence = (-sequence)")
+  end
+
+  def last_sequence
+    self.class.last&.sequence.to_i
   end
 end
--- a/app/models/workflow.rb
+++ b/app/models/workflow.rb
@@ -58,23 +58,26 @@ class Workflow < ApplicationRecord
 
   # sequences increment between [sequence sequence_was sequence)
   def sequence_higher(startp, endp = nil)
-    query = self.class.where(table[:sequence].gteq(startp))
-    query = query.where(table[:sequence].lteq(endp)) if endp
-    query.update_all("sequence = (-sequence - 1)")
-
-    query = self.class.where(table[:sequence].lt(0))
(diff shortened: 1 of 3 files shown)


app/models/workflow.rb
```diff
@@ -29,16 +30,54 @@ def table
...
-  # move all related sequence number one higher if the desired number is in use
-  def move_rest_higher
-    return unless sequence && sequence_changed? && self.class.exists?(:sequence => sequence)
+  # no gap between sequences
+  def adjust_sequences
+    return unless sequence_changed?
 
-    self.class.where(table[:sequence].gteq(sequence)).update_all("sequence = (-sequence - 1)")
-    self.class.where(table[:sequence].lt(0)).update_all("sequence = (-sequence)")
+    throw :abort if sequence && sequence <= 0
```
mkanoor, line 47: @bzwei Should this raise an InvalidArguments Exception so the user knows that they have supplied an invalid parameter

bzwei: I learned this from rails doc. `throw :abort` will guide rails to raise an `RecordInvalid` exception.

spec/models/workflow_spec.rb
```diff
@@ -22,39 +22,69 @@
...
+    end
+
+    it 'moves a workflow to the top' do
+      old_ids = Workflow.pluck(:id)
+      Workflow.find(old_ids[3]).update(:sequence => 1)
+      expect(Workflow.pluck(:id)).to eq([old_ids[3], old_ids[0], old_ids[1], old_ids[2], old_ids[4]])
+      expect(Workflow.pluck(:sequence)).to eq([1, 2, 3, 4, 5])
+    end
+
+    it 'moves a worflows to the bottom' do
```
eclarizio, line 51: "workflow" typo here

spec/models/workflow_spec.rb
```diff
@@ -22,39 +22,69 @@
...
 
-    context 'when sequence has new value' do
-      it 'updates with the new sequence' do
-        workflow.update(:sequence => old_sequence + 1)
-        workflow.reload
-        expect(workflow.sequence).to eq(old_sequence + 1)
-      end
+    it 'fails to update non-positive sequence' do
+      expect { Workflow.first.update!(:sequence => 0) }.to raise_error(ActiveRecord::RecordInvalid)
     end
```
eclarizio, line 81: I think if you want to keep this spec since it's just a duplicate of 40-42, the description should change to something like "it fails to update with a 0 sequence".

app/models/workflow.rb
```diff
@@ -29,16 +30,54 @@ def table
...
+    else
+      sequence_higher(sequence, sequence_was)
+    end
+  end
+
+  # sequences increment between [sequence sequence_was sequence)
+  def sequence_higher(startp, endp = nil)
+    query = self.class.where(table[:sequence].gteq(startp))
+    query = query.where(table[:sequence].lteq(endp)) if endp
+    query.update_all("sequence = (-sequence - 1)")
```
eclarizio, line 63: Do you think it makes sense to extract this block out (as well as 72-74) into a separate method that has a good name of what this is doing and pass in the `-1` from 63. And then for 72-74 it could also use this new method, but pass in `1` instead. Small "DRY".

spec/support/tag_resources_shared_context.rb
```diff
@@ -1,11 +1,12 @@
...
   let(:tenant2) { create(:tenant, :external_tenant => 'fred') }
-  let(:workflow1) { create(:workflow, :tenant => tenant1, :sequence => 10) }
+  let(:tenant1_workflows) { create_list(:workflow, 2, :tenant => tenant1) }
+  let(:workflow1) { tenant1_workflows[1] }
   let(:wf1_tag)   { "/approval/workflows=#{workflow1.id}" }
-  let(:workflow2) { create(:workflow, :tenant => tenant1, :sequence => 5) }
+  let(:workflow2) { tenant1_workflows[0] }
   let(:wf2_tag)   { "/approval/workflows=#{workflow2.id}" }
-  let(:workflow3) { create(:workflow, :tenant => tenant2, :sequence => 7) }
+  let(:workflow3) { create(:workflow, :tenant => tenant2) }
```
eclarizio, line 9: Were these just necessary changes due to the `before_validation` changes? I don't see any use of `workflow1/2/3` anywhere in files you touched so that's why I'm curious.

bzwei: Because of this PR, hard coded sequence number is not honored but depend on the order of workflow creations. So I introduced `tenant1_workflows` to ensure `workflow1` is before `workflow2`. `workflow/1/2/3` are used in https://github.com/RedHatInsights/approval-api/blob/master/spec/services/workflow_find_service_spec.rb#L45

bzwei:
@eclarizio We will wait extracting the common logic until there is a similar need somewhere else. 

 I think the SQL Injection alarm is a false one because there is no injection in the query statement.

gmcculloug:
@bzwei I think we do need to address the Hakiri warning or that will not go away.  They are reviewing the code base, no just the PR changes.  I can help here if you want.

bzwei:
@gmcculloug Yes I need help. The inject like syntax is in the update_all statement, not the query statement. There was no warning before. Once I move the same code into a separate method as @eclarizio recommends, the warning appears. I think it is a false alarm. Hakiri says we can either fix it or mark it false alarm. I don't see where I can mark it.

app/models/workflow.rb
```diff
@@ -29,16 +30,57 @@ def table
...
+  # sequences decrement between [startp endp]
+  def sequence_lower(startp = sequence, endp = nil)
+    change_sequences_to_negative(startp, endp, 1)
+    change_sequences_to_positive(-startp + 1)
+  end
+
+  def change_sequences_to_negative(startp, endp, delta)
+    query = self.class.where(table[:sequence].gteq(startp))
+    query = query.where(table[:sequence].lteq(endp)) if endp
+    query.update_all("sequence = (-sequence + (#{delta}))")
```
gmcculloug, line 74: @bzwei Change this to 
```ruby
query.update_all(["sequence = (-sequence + (?))", delta])
```

gmcculloug:
@mkanoor PTAL, would like to get your approval before merging.

gmcculloug:
Thanks @mkanoor The original UI goal was to have the UI allow for drag & drop functionality here but the components were not available to support that which required us to expose the sequence so an admin could modify them.

@sbuenafe-rh @Hyperkid123 I believe this component situation is still the same, correct?

Either way, I think is this ready to merge and we can continue to improve the UI in this area.

mkanoor:
@gmcculloug I am wondering if there is another way of implementing the same thing. A user selects one or more workflows (ID1, ID3) and moves them somewhere else
- Top.  (:top)
- Bottom (last:)
- Between two workflows (ID55:ID46)

Then the algorithm becomes like drop a collection between 2 book ends and we make room for the dropped pieces by adjusting the sequences of where something got dropped. This way we don't have to pass the sequences around

The API would get 2 parameters
ids : Array of Workflow IDs being moved
location: The ID of the Workflow where the collection ids need to be placed with special values like top/end/actual_workflow_id

Hyperkid123:
> @sbuenafe-rh @Hyperkid123 I believe this component situation is still the same, correct?

There is no direct PF support for DnD components. However, I don't think that will happen or that it should ever happen. In React, usually, you use a third-party DnD library which only provides the DnD API and logic and applies that to your components. So if we want DnD, we can implement it using PF (same as with the survey editor which has the DnD feature disabled in catalog).

PR #357, opened 2020-05-18
[Stable] Adjust sequences only when current value changes

gmcculloug:
Stable branch PR of https://github.com/RedHatInsights/approval-api/pull/353
Also includes PR https://github.com/RedHatInsights/approval-api/pull/351 `Switched to use the insights-rbac-api-client gem`

> We should not adjust sequences if current value does not change. Otherwise each save causes the numbers moving higher.
> 
> https://projects.engineering.redhat.com/browse/SSP-1536
> 
> The above ticket reported the problem occurs at GET. It is because each get does validation on groups and save with updated group names.
> 
> It also resolves a warning on deprecated call.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -18,7 +18,6 @@ gem 'rails',               '>= 5.2.2.1', '~> 5.2.2'
 gem 'sprockets',           '~> 3.7.2'
 
 gem 'kie_client', :git => "https://github.com/RedHatInsights/kie-api-client-ruby", :branch => "master"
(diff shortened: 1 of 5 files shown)


gmcculloug:
@bzwei Needed to include https://github.com/RedHatInsights/approval-api/pull/351 as well based on the gem name change.

gmcculloug:
@zsadeh smoke-tests failed on the first try so I restarted.  Just wanted to make sure there were no known issues you are aware of.  Maybe just resourcing at this time of day.

zsadeh:
@gmcculloug it is deployment issues - I suggest to merge it and I will test it on QA env ASAP

PR #358, opened 2020-05-19
Retry building openapi-generator jar file

bzwei:
Occasionally travis validation of openapi fails at building the openapi-generator-cli.jar. With this enhancement if the first validation attempt fails we will check whether the jar file exists. If not, we will start at re-downloading the script and run it again (retry once).

### DIFF
--- a/.travis_scripts/openapi-validator.sh
+++ b/.travis_scripts/openapi-validator.sh
@@ -1,7 +1,14 @@
 #!/usr/bin/env bash
-set -e
+set -eu
+set -o pipefail
+
+if ! ./openapi-generator-cli version
+then
+  echo "openapi-generator-cli version failed, likely caused by a mvn build failure. Retry once."
+  ./openapi-generator-cli version
+fi
+
 for entry in ./public/doc/openapi-3-v*.json
 do
   ./openapi-generator-cli validate -i "$entry"
 done
-


.travis_scripts/openapi-validator.sh
```diff
@@ -1,7 +1,25 @@
 #!/usr/bin/env bash
-set -e
+set -eu
+set -o pipefail
+
+function download_generator {
+  curl -L https://raw.githubusercontent.com/OpenAPITools/openapi-generator/master/bin/utils/openapi-generator-cli.sh -o ./openapi-generator-cli
+  chmod +x ./openapi-generator-cli
```
mkanoor, line 7: @bzwei Why not have the retry logic here
you could check $? to get the exit code of the curl command and then retry the download again.
I am not sure if we have to to run the validate in order to figure out that the download failed and then issue a retry.

bzwei: The download normally has no problem. Running the command line may have the exit code nonzero either because the true validation fails (desired) or the maven build fails (undesired, retry).

bzwei: The script first build a jar file and then run the jar to validate openapi.

.travis_scripts/openapi-validator.sh
```diff
@@ -1,7 +1,25 @@
...
-  ./openapi-generator-cli validate -i "$entry"
+  if ! ./openapi-generator-cli validate -i "$entry"
+  then
+    if ls openapi-generator-cli*.jar 1>/dev/null 2>&1
+    then
+      exit 1
+    else
+      echo "openapi-generator-cli jar file build failed. Retry..."
+      download_generator
+      ./openapi-generator-cli validate -i "$entry"
```
mkanoor, line 21: @bzwei According to the comments in the script 

# This script will query github on every invocation to pull the latest released version
# of openapi-generator.

The failure is in the maven repo so why are we refetching the initial script.
Shouldn't we just have to run the 
```
./openapi-generator-cli version
```
Multiple times to make sure if can download the jar from the maven repo. Once downloaded it can be reused and if we cache the version using

```
OPENAPI_GENERATOR_VERSION=`openapi-generator version`
```
Then we can run all the validation using the latest version which will help when the version switch happens between your invocations. If the version command fails you can retry the version command twice till you get a successful return code.

bzwei: @mkanoor We do not need to download it multiple times. I did it just in case the file may be corrupted. I can remove it and only re-run the script.

eclarizio: Do we not need to do another `ls openapi-generator-cli*.jar 1>/dev/null 2>&1` here to check if this worked or not? If it doesn't, how will we know?

mkanoor: @bzwei @eclarizio 
We need the file to be complete, if its a partial or corrupt file the "ls" is going to succeed.
When we execute the version command with it we know that the whole jar is there and is valid and it can execute commands. If the version command fails we can do a retry once.

gmcculloug:
@bzwei This change should be applied on the other repos where we added the validation check, right?  I just saw a failure on the catalog-api related to this.

bzwei:
@gmcculloug yes. I will do it.

PR #359, opened 2020-05-21
Set rbac call pagenation limit to 500 where applicable

bzwei:
https://projects.engineering.redhat.com/browse/SSP-1531


hsong-rh:
@mkanoor This is the one we discussed before, may need RBAC to enhance: return access list for group

mkanoor:
@bzwei @bdunne In the process to do validation we are running a logic something similar to this

```
For each group_uuids 
   GET /groups/{group_uuid}/roles/
   For each roles
        GET /roles/{role_uuid}/access/
```

With this approach we are going to be fetching the same role multiple times if they are assigned to  different groups.
e.g
group1 -> Cost Admin Role , group2 -> Cost Admin Role, group3 -> Approval Approver Role
we will fetch the Cost Admin role twice
If we just cached the roles as we are traversing each group we dont have to refetch it.

mkanoor:
I have created a JIRA ticket for the RBAC team to implement this on their side
https://projects.engineering.redhat.com/browse/RHCLOUD-6611

hsong-rh:
@mkanoor yes, for multiple groups scenario duplicate roles will be processed twice. I created an issue: https://github.com/RedHatInsights/approval-api/issues/360, will enhance this later

bzwei:
@mkanoor is everything good for this PR, other than the Jira to be resolved by the rbac team?

PR #362, opened 2020-05-22
Add checking on state of requests before deleting approval process

hsong-rh:
In this PR, workflow is forbid to delete if one of its associated requests is still running: (state is `pending`, `started` or `notified`). Try to delete such a workflow will raise an error. 

Workflow will allow to be deleted If all its associated requests are completed. After workflow is deleted, the foreign key `workflow_id` of these completed requests will set to `nil`.

Whenever the workflow is deleted, the associated `TagLinks` will always be removed too. 

https://projects.engineering.redhat.com/browse/SSP-1544

### DIFF
--- a/app/models/workflow.rb
+++ b/app/models/workflow.rb
@@ -5,7 +5,7 @@ class Workflow < ApplicationRecord
   default_scope { order(:sequence => :asc) }
 
   belongs_to :template
-  before_destroy :validate_deletable, :pretend => true
+  before_destroy :validate_deletable, :prepend => true
   has_many :requests, -> { order(:id => :asc) }, :inverse_of => :workflow, :dependent => :nullify
   has_many :tag_links, :dependent => :destroy, :inverse_of => :workflow
 
--- a/app/models/workflow.rb
+++ b/app/models/workflow.rb
@@ -6,6 +6,7 @@ class Workflow < ApplicationRecord
 
   belongs_to :template
   before_destroy :validate_deletable, :prepend => true
+  after_commit :send_deletion_message
   has_many :requests, -> { order(:id => :asc) }, :inverse_of => :workflow, :dependent => :nullify
   has_many :tag_links, :dependent => :destroy, :inverse_of => :workflow
 
@@ -47,6 +48,10 @@ class Workflow < ApplicationRecord
         dependencies[key] << value
       end
     end
+  end
+
+  def send_deletion_message
+    EventService.new(nil).workflow_deleted(id)
   end
 
   def table
--- a/app/models/workflow.rb
+++ b/app/models/workflow.rb
@@ -5,7 +5,8 @@ class Workflow < ApplicationRecord
   default_scope { order(:sequence => :asc) }
 
   belongs_to :template
-  has_many :requests, -> { order(:id => :asc) }, :inverse_of => :workflow
+  before_destroy :validate_deletable, :pretend => true
+  has_many :requests, -> { order(:id => :asc) }, :inverse_of => :workflow, :dependent => :nullify
   has_many :tag_links, :dependent => :destroy, :inverse_of => :workflow
 
   validates :name, :presence => true, :uniqueness => {:scope => :tenant}
@@ -29,7 +30,15 @@ class Workflow < ApplicationRecord
     super.merge(:object_dependencies => object_dependencies)
   end
 
+  def deletable?
+    requests.any? { |request| !request.finished? } ? false : true
+  end
+
   private
+
+  def validate_deletable
+    throw :abort unless deletable?
+  end
 
   def object_dependencies
     {}.tap do |dependencies|
--- a/app/policies/workflow_policy.rb
+++ b/app/policies/workflow_policy.rb
@@ -16,7 +16,7 @@ class WorkflowPolicy < ApplicationPolicy
   end
 
   def destroy?
-    permission_check('delete')
+    permission_check('delete') ? record.deletable? : false
   end
 
   def link?
--- a/app/services/event_service.rb
+++ b/app/services/event_service.rb
@@ -6,6 +6,7 @@ class EventService
   EVENT_REQUEST_CANCELED  = 'request_canceled'.freeze
   EVENT_APPROVER_GROUP_NOTIFIED = 'approver_group_notified'.freeze
   EVENT_APPROVER_GROUP_FINISHED = 'approver_group_finished'.freeze
+  EVENT_WORKFLOW_DELETED = 'workflow_deleted'.freeze
   EVENT_SENDER = 'approval_service'.freeze
 
   attr_accessor :request
@@ -32,6 +33,10 @@ class EventService
                :group_name => request.group_name,
                :decision   => request.decision,
                :reason     => request.reason || '')
+  end
+
+  def workflow_deleted(workflow_id)
+    send_event(EVENT_WORKFLOW_DELETED, :workflow_id => workflow_id)
   end
 
   # request is root
--- a/spec/models/workflow_spec.rb
+++ b/spec/models/workflow_spec.rb
@@ -88,6 +88,71 @@ RSpec.describe Workflow, :type => :model do
     end
   end
 
+  describe '#deletable?' do
+    shared_examples_for "undeletable_states" do |state|
+      it "returns false for #{state}" do
+        request = create(:request, :workflow => workflow, :state => state)
+        expect(workflow.deletable?).to eq(false)
+      end
+    end
+    
+    shared_examples_for "deletable_states" do |state|
+      it "returns true for #{state}" do
+        request = create(:request, :workflow => workflow, :state => state)
+        expect(workflow.deletable?).to eq(true)
+      end
+    end
+
+    (Request::STATES - Request::FINISHED_STATES).each do |state|
+      it_behaves_like "undeletable_states", state
+    end
+
+    Request::FINISHED_STATES.each do |state|
+      it_behaves_like "deletable_states", state
+    end
+  end
+
+  describe '#destroy' do
+    let!(:taglink) { create(:tag_link, :workflow => workflow) }
+
+    context 'when without associated request' do
+      it 'deletes workflow' do
+        expect(TagLink.count).to eq(1)
+        workflow.destroy
+
+        expect(workflow).to be_destroyed
+        expect(TagLink.count).to eq(0)
+      end
+    end
+
+    context 'when associated with requests' do
+      it "is not deletable" do
+        allow(workflow).to receive(:deletable?).and_return(false)
+
+        request = create(:request, :workflow => workflow)
+
+        expect(TagLink.count).to eq(1)
+        expect { workflow.destroy! }.to raise_error(ActiveRecord::RecordNotDestroyed)
+        expect(workflow).not_to be_destroyed
+        expect(request.workflow_id).to eq(workflow.id)
+        expect(TagLink.count).to eq(1)
+      end
+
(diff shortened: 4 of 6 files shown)


app/models/request.rb
```diff
@@ -74,6 +74,10 @@ def finished?
     FINISHED_STATES.include?(state)
   end
 
+  def running?
```
bzwei, line 77: do not need this method. You can use `!finished?`

app/models/concerns/approval_states.rb
```diff
@@ -7,5 +7,6 @@ module ApprovalStates
   CANCELED_STATE  = 'canceled'.freeze
   FAILED_STATE    = 'failed'.freeze
   STATES = [PENDING_STATE, SKIPPED_STATE, STARTED_STATE, NOTIFIED_STATE, COMPLETED_STATE, CANCELED_STATE, FAILED_STATE].freeze
+  RUNNING_STATES = [STARTED_STATE, NOTIFIED_STATE].freeze
```
bzwei, line 10: no need. Technically pending_state cannot be deleted either.

app/models/workflow.rb
```diff
@@ -25,6 +26,10 @@ def external_signal?
 
   private
 
+  def cleanup_requests_and_taglinks
```
bzwei, line 29: why name the method this way? There is no actual cleanup of requests and tag links happening here. You may name it something like `validate_deletable`

bzwei:
@hsong-rh When a workflow is deleted, should we delete remote tags according to the tag link? All are we even able to delete the remote tags since we don't have the object_id.

Secondly you need to provide an accurate capability for each workflow telling whether it is deletable.

hsong-rh:
@bzwei We cannot delete remote tags if `object_id` is not available.

bzwei:
ping @mkanoor

app/models/workflow.rb
```diff
@@ -5,7 +5,8 @@ class Workflow < ApplicationRecord
   default_scope { order(:sequence => :asc) }
 
   belongs_to :template
-  has_many :requests, -> { order(:id => :asc) }, :inverse_of => :workflow
+  before_destroy :validate_deletable, :pretend => true
```
gmcculloug, line 8: Did you mean `:prepend => true` ?

https://stackoverflow.com/questions/49577447/rails-5-before-destroy-callback-with-prepend-or-without-fails

hsong-rh: Yes! 👍

gmcculloug:
@bzwei @mkanoor Is the proper fix here to post a Kafka message that the object was deleted that the other service need to be taught to action on?

bzwei:
@gmcculloug What you proposed can be a solution. Not deleting the remote tag is not a critical problem. It leaves the tag remains useless. With the kafka notification at least we give the other service a chance to cleanup.

hsong-rh:
@bzwei @gmcculloug The logic to broadcast Kafka message is added.

app/models/workflow.rb
```diff
@@ -23,7 +24,19 @@ def external_signal?
...
+    requests.any? { |request| !request.finished? } ? false : true
+  end
+
   private
+  
+  def validate_deletable
+    throw :abort unless deletable?
+
+    requests.select { |request| request.finished? }.each do |request|
+      EventService.new(request).workflow_deleted
```
gmcculloug, line 37: I'm confused by a few things here:
1. The `deletable?` method is already checking for finished request and we would not get here if there are still pending requests.  So do we need the select loop at all?  (More in this loop in the next item)
2. I feel like the event should be a life-cycle event on the workflow object and I would not expect a separate event per-request it is linked to.  Meaning, I was expecting a single "workflow deleted" event message.
3. If someone needs to know that the workflow was removed from the completed request (not sure we have a use-case right now) I would expect that to be multiple events against the request object.

cc @mkanoor @eclarizio

gmcculloug:
@bzwei To followup on the discussion this morning, we are trying to provide useful statistics when listing object in the catalog api.  One of the values asked for was a way to determine if the Portfolio (and eventually other objects) have Approvals assigned.

I was working on the idea that we could determine this without making calls to external services by checking the tag assignments. (See https://github.com/RedHatInsights/catalog-api/pull/700/files#diff-9856846af9bff5bb7ca920879c4d430eR39)

But the conversation in this PR made me realize that if the Approval process is deleted from the Approval API, Catalog would not know to remove the tag.  As you pointed, the overall Approval logic would continue to work as expected since we would pass the tag to approval and it would not match any process.  However, this means the Catalog would believe it has a Approval process assigned without doing an external API call to validate.  I'm removing the `approval_processes` data in the initial PR but I do intend to bring that back at some point.

app/models/workflow.rb
hsong-rh: If a `workflow` is deletable, we only update its associated `requests`, nullify their `workflow_id`, not delete them. I will change to broadcast single message in this case.

app/models/workflow.rb
```diff
@@ -23,7 +24,17 @@ def external_signal?
...
+  def deletable?
+    requests.any? { |request| !request.finished? } ? false : true
+  end
+
   private
+  
+  def validate_deletable
+    throw :abort unless deletable?
+
+    requests.any? { EventService.new(requests.first).workflow_deleted }
```
bzwei, line 36: We should use an after_destroy callback instead of piggybacking here in a before callback.

In the event service we may allow a null request in the constructor and pass the `workflow_id` in the argument list of `workflow_deleted` method

mkanoor:
@hsong-rh @gmcculloug @lgalis 
There is another piece related to deleting the workflow. The UI is going to show nothing if the previously assigned workflow to a portfolio gets deleted. Should the UI be raising a warning saying that the workflow is missing.

bzwei:
Follow @mkanoor's comment. When the user sees the warning (BTW, we probably would say the workflow no longer exist rather than missing), he likely wound go (or we encourage him) to unlink the workflow, but API call will fail.

We could enhance the unlink api to accept a no longer existing workflow_id, and remove the remote tag.

app/models/workflow.rb
```diff
@@ -5,7 +5,9 @@ class Workflow < ApplicationRecord
   default_scope { order(:sequence => :asc) }
 
   belongs_to :template
-  has_many :requests, -> { order(:id => :asc) }, :inverse_of => :workflow
+  before_destroy :validate_deletable, :prepend => true
+  after_destroy :send_deletion_message
```
bzwei, line 9: should change to `after_commit`

hsong-rh: yes, make sure the message is only broadcasted after AR is really deleted

gmcculloug:
@mkanoor @bzwei I have been think about this issue a lot recently.  The issue of deleting the Approval Process from the Approval API currently will lead to the tag existing on the resource that assigned it.  This caused me to remove the approval_process count from the Catalog metadata logic as there was no easy way to guarantee the accuracy of the data.

I believe there are two approaches here and arguments to be made for both:

1. The deletion raises a Kafka event (as of this PR) and the individual applications can monitor and react to the event.  My initial thought was they should just remove the tag.  It is a clean micro-service solution.  But for Approvals maybe causes an issue if something can now be ordered without an approval unknowingly.

2. Monitor for the event and flag the object somehow saying that the Approval process was deleted.  I'm not really sure what this means at the moment, so would have to be designed out.  But imagine we capture some record about the deletion and an Catalog objects effected by it would need to be reviewed/cleared by a Catalog Admin before the objects could be ordered.

In either case once the delete is initiated from the Approval API the other services are left with a tag that only identifies an Approval process ID.  Not very useful after deletion.

My suggestion at the moment is that we open an issue to track this work and discuss with PM.

hsong-rh:
Changed to `wip`

bzwei:
@gmcculloug It is my concerning too that an order can go without approval because of a deleted approval process(workflow). In approval-api side we may have the following options:
1. Do not allow a workflow to be deleted if it has tag_links associations, i.e., if it has been linked to any type of external resources. It also means the workflow can never be deleted. We may consider to add an API to delete tag_links, or
2. Assume we will always have internal tag for approvals. If we found an approval request has internal approval tag but the workflow has been deleted, we raise an error. It cannot handle the case  if in the future we can link an external tag with an approval workflow, or
3. Never delete tag_links even if the workflow is deleted. Raise an error if we can find a matched tag_link record with request's external resources but the workflow is already deleted. It also has the down side that tag_links remain in the database.

We should have a meeting to discuss workflow deletion.

mkanoor:
There are 2 aspects here

1. Are we using statically linking the resource and workflow object. Which is what the current implementation is.
2. Are we using dynamic linking where the resource and the workflow have no connection other than having some shared tag.

Some of our current implementation is a mix on 1 and 2. We might have to decide which solution we want to lean towards. If we want to use 1 (static linking) we had discussed that we should be adding the ID of the remote object in the taglink object that will allow us to get the reference to the remote object.
With 2 we won't know at display time what workflows are connected to a collection of objects in a job (portfolios, inventory, sources) We would always need to have a dry run to collect the tags of all related objects and ask the Approval Service which approval processes are involved given the collection of tags. This could also be done on a single object by object basis (portfolio, source, inventory)

bzwei:
We may want to write `WORKFLOW_DELETED` messages to a separate topic, so that the subscriber such as topology will not receive other messages such as `REQUEST_APPROVED` which are not to its interest.

Currently approval-api writes all messages to `platform.approval`, if we name the new topic such as `platform.approval.dependency`, catalog-api can subscribe to both topics using a regex.

gmcculloug:
Need some help evaluating @bzwei previous comment (https://github.com/RedHatInsights/approval-api/pull/362#issuecomment-644871855) on creating a new topic.  I am not sure what the expected behavior would be here for a service subscribed to this topic.d

@mkanoor do you have any thoughts on this?

mkanoor:
@bzwei The Remote objects that are tagged can be in Topology/Sources/Catalog so does it make sense to keep the Delete Message in a separate topic. Why can't it be in a single topic since the listener can decide what to do based on the message

gmcculloug:
Was have been thinking about this as well, and would probably vote to keep the same topic unless we had a good reason.  We can always split it out later if we find the need to separate them.

@bzwei What do you think?

@syncrou Any concerns with the current approach for Topology/Sources?

syncrou:
@gmcculloug I certainly would rather keep the same topic vs having to manage multiple. It sounds like between @mkanoor's comment [here](https://github.com/RedHatInsights/approval-api/pull/362#issuecomment-647560097) and @gmcculloug [here](https://github.com/RedHatInsights/approval-api/pull/362#issuecomment-647639609) that if a single topic is problematic we can reevaluate after.

bzwei:
Sure. Since majority vote to have one single topic, let's keep this way.

gmcculloug:
@mkanoor PTAL

app/models/workflow.rb
```diff
@@ -29,8 +31,16 @@ def metadata
...
   end
 
+  def deletable?
+    requests.any? { |request| !request.finished? } ? false : true
+  end
+
   private
 
+  def validate_deletable
+    throw :abort unless deletable?
```
mkanoor, line 41: @hsong-rh Why this a throw and not a raise exception. Do callbacks suppress exceptions. If we can have a "WorkflowIsActiveException" we would know why exactly a failure happened.

hsong-rh: It's recommended by rails doc (https://guides.rubyonrails.org/active_record_callbacks.html#halting-execution) to stop the chain.

PR #367, opened 2020-06-03
Prevent workflow sequence from becoming negative

bzwei:
Validate sequence to be all positive after save or destroy. Introduce `WorkflowDeleteService` class.

https://projects.engineering.redhat.com/browse/SSP-1567

It as been reported that the workflow sequence sometimes can become negative. It can be reproduced by simultaneously deleting multiple workflows.

Before updating or destroying a workflow, we adjust sequences for other workflows and leave the current workflow's sequence number to be negative and allow it to be updated or destroyed through the AR's built-in process. But in a concurrent situation we may identify the wrong sequence number and leave a negative sequence un-updated.

The solution is to add validation in the `after_save` or `after_destroy` callbacks to ensure no negative sequence number is ever allowed. Transaction rolls back if the validation fails. We then retry the operation using the reloaded record.

### DIFF
--- a/app/services/workflow_delete_service.rb
+++ b/app/services/workflow_delete_service.rb
@@ -1,0 +1,17 @@
+class WorkflowDeleteService
+  attr_accessor :workflow_id
+
+  def initialize(workflow_id)
+    self.workflow_id = workflow_id
+  end
+
+  def destroy
+    begin
+      retries ||= 0
+      Workflow.find(workflow_id).destroy!
+    rescue ActiveRecord::RecordNotUnique, Exceptions::NegativeSequence # Failed to update sequence after deletion due to concurrent issue
+      retry if (retries += 1) < 3
+    end
+  end
+
+end
--- a/lib/exceptions.rb
+++ b/lib/exceptions.rb
@@ -6,4 +6,5 @@ module Exceptions
   class KieError < StandardError; end
   class InvalidStateTransitionError < StandardError; end
   class UserError < StandardError; end
+  class NegativeSequence < StandardError; end
 end
--- a/app/controllers/api/v1x2/workflows_controller.rb
+++ b/app/controllers/api/v1x2/workflows_controller.rb
@@ -34,7 +34,7 @@ module Api
         workflow = Workflow.find(params.require(:id))
         authorize workflow
 
-        workflow.destroy!
+        WorkflowDeleteService.new(params[:id]).destroy
         head :no_content
       rescue ActiveRecord::InvalidForeignKey => e
         json_response({ :message => e.message }, :forbidden)
--- a/app/models/workflow.rb
+++ b/app/models/workflow.rb
@@ -13,7 +13,9 @@ class Workflow < ApplicationRecord
 
   before_validation :new_sequence, :on => :create
   before_validation :adjust_sequences, :on => :update
+  after_save        :validate_positive_sequences
   before_destroy    :sequence_lower
+  after_destroy     :validate_positive_sequences
 
   def external_processing?
     template&.process_setting.present?
@@ -82,5 +84,9 @@ class Workflow < ApplicationRecord
 
   def last_sequence
     self.class.last&.sequence.to_i
+  end
+
+  def validate_positive_sequences
+    raise Exceptions::NegativeSequence if self.class.where(table[:sequence].lteq(0)).exists?
   end
 end
--- a/app/services/workflow_update_service.rb
+++ b/app/services/workflow_update_service.rb
@@ -2,10 +2,10 @@ require_relative 'mixins/group_validate_mixin'
 
 class WorkflowUpdateService
   include GroupValidateMixin
-  attr_accessor :workflow
+  attr_accessor :workflow_id
 
   def initialize(workflow_id)
-    self.workflow = Workflow.find(workflow_id)
+    self.workflow_id = workflow_id
   end
 
   def update(options)
@@ -14,9 +14,9 @@ class WorkflowUpdateService
     end
 
     begin
-      retries ||=0
-      workflow.update!(options)
-    rescue ActiveRecord::RecordNotUnique # Sequence numbers may be found duplicated due to concurrent issue
+      retries ||= 0
+      Workflow.find(workflow_id).update!(options)
+    rescue ActiveRecord::RecordNotUnique, Exceptions::NegativeSequence # Sequence numbers may be found duplicated due to concurrent issue
       retry if (retries += 1) < 3
     end
   end
--- a/config/initializers/custom_exception_mappings.rb
+++ b/config/initializers/custom_exception_mappings.rb
@@ -2,8 +2,10 @@
 ActionDispatch::ExceptionWrapper.rescue_responses.merge!(
   "ActiveRecord::RecordNotSaved"               => :bad_request,
   "ActiveRecord::RecordInvalid"                => :bad_request,
+  "ActiveRecord::RecordNotUnique"              => :bad_request,
   "ActionController::ParameterMissing"         => :bad_request,
   "Exceptions::InvalidStateTransitionError"    => :bad_request,
+  "Exceptions::NegativeSequenceError"          => :bad_request,
   "Exceptions::UserError"                      => :bad_request,
   "Exceptions::NotAuthorizedError"             => :forbidden,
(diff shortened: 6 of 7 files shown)


lib/exceptions.rb
```diff
@@ -6,4 +6,5 @@ class RBACError < StandardError; end
   class KieError < StandardError; end
   class InvalidStateTransitionError < StandardError; end
   class UserError < StandardError; end
+  class NegativeSequenceError < StandardError; end
```
hsong-rh, line 9: Even with retry, this error still has chance to be raised. Should add it in the mapping: https://github.com/RedHatInsights/approval-api/blob/master/config/initializers/custom_exception_mappings.rb#L2

app/services/workflow_delete_service.rb
```diff
@@ -0,0 +1,17 @@
+class WorkflowDeleteService
+  attr_accessor :workflow_id
+
+  def initialize(workflow_id)
+    self.workflow_id = workflow_id
+  end
+
+  def destroy
```
gmcculloug, line 8: @bzwei @hsong-rh I feel like we need to re-sync on how we are defining service classes in this app.  At the start of the Catalog/Approval work we had talked about keeping services uniform following a structure which you can see here: https://github.com/RedHatInsights/catalog-api/blob/master/app/services/api/v1x0/catalog/add_to_order.rb

Basically, a `initialize` method if needed and a `process` method that always returned `self`.  Plus, I don't believe the service classes need to have "Service" in the name as the class names would normal define the action they were performing.  (Like the `AddToOrder` class above.) It looks like we have drifted away from this concept but I feel like it is something we should try to get back into alignment and uniform across the projects.  I'm sorry that I failed to catch this in earlier PRs.

@eclarizio Curious if you have thoughts on this topic.

app/services/workflow_create_service.rb
```diff
@@ -14,7 +14,7 @@ def create(options)
     begin
       retries ||= 0
       template.workflows.create!(options)
-    rescue ActiveRecord::RecordNotUnique # The auto generated sequence number may be found duplicated due to concurrent issue
+    rescue ActiveRecord::RecordNotUnique, Exceptions::NegativeSequenceError # The auto generated sequence number may be found duplicated due to concurrent issue
       retry if (retries += 1) < 3
```
gmcculloug, line 18: @bzwei I get that the model will raise the new `NegativeSequenceError` message on create, update or destroy but how does retrying potentially fix the situation?

And when we run out of retries it means we need a manual process to correct it?  Did the gem we were using before have this same exposure?

lib/exceptions.rb
gmcculloug, line 9: Recommend using just `NegativeSequence`

app/services/workflow_create_service.rb
bzwei: True. It is not necessary to catch `NegativeSequenceError` for create. I added to make the same as how we catch similar errors in other places. I can remove it.

lib/exceptions.rb
bzwei: It is a personal preference. I tend to append Error in the end so that we can clearly know it is an error class different from regular class, just like in the Java world every exception class ends with `Exception`. I can remove the trailing world.

app/services/workflow_create_service.rb
gmcculloug: My question is really more about the logic in general.

1. How does the retry fix the issue?
2. How do we fix this when we run out of retires and the issue still exists?
3. Did the gem have the same exposure?

app/services/workflow_delete_service.rb
bzwei: @gmcculloug We did not make all service class to implement common `process` method from day one, so it is not really a drifting away. Catalog/approval are technically two independent projects, we are thinking we can take some different approaches.

Our current service classes have their own naming conventions. All classes end with `Service` to indicate they are service classes, just like controller classes end with `Controller`. Each class has verbs to indicate what function the method is about to perform. We think `process` is too generic. More importantly we allow one service to perform multiple closely related operations. BTW, if each service class only have one `process`, what is the use of the returned `self`?

Above is our reasoning. We are willing to move to `process` style if you all think it is more appropriate.

app/services/workflow_create_service.rb
bzwei: The gem we used before did not allow to make `sequence` unique. So it does raise an error but may result in bad record.
If retry does not solve the problem (should be very rare), then yes, we will fail the api so the user can try later. But we should not need a manual fix in the database.

lib/exceptions.rb
eclarizio: I think that adding `Error` to the end depends on the actual error. For example if it's another process that is having the problem, then it makes sense, like `TopologyError` or in the case of these here, `KieError`, or `RBACError`. But if it's a state that we came up with, I think dropping the `Error` suffix makes sense. `InvalidStateTransition`, `NegativeSequence`. 

This follows what we're doing in catalog with stuff like `InvalidSurvey`, `OrderUncancelable`, but we have `SourcesError` and `TopologyError` and `ApprovalError`.

I don't have a super strong opinion on this, I could go either way, just pointing out how we currently handle these.

app/services/workflow_delete_service.rb
eclarizio: @gmcculloug @bzwei I don't particularly like the whole `process` that always returns `self`, but it *is* a pattern that we have followed for a while so that's kind of the only reason I can see to continue it. I think that if we ever go to some sort of service delegator or something that calls the various services it makes it easy to handle since all the services are homogenous with their main method.

In the past, I've generally done as Bill mentioned, suffixed the class with a `Service`. As I think about it more, though, it does feel like adding the `Service` suffix is a bit redundant, given that the classes themselves are in the `app/services` folder. I can go either way on adding `Service` as a suffix, since it does basically follow rail convention with controllers, and helpers, though models obviously don't have `Model` as a suffix.

I think the argument that `process` is too generic is sound, but on the other hand I think if you're limiting your service to a single responsibility (which we should be, to keep them simple), the name of the service itself should be indicative enough of what it does, and `process` is fine. You wouldn't expect the `WorkflowDeleteService` to go and do anything other than delete workflows, so `WorkflowDeleteService.new.delete` is also redundant.

> "More importantly we allow one service to perform multiple closely related operations"

I think as silly as it might seem, separating these things out into two separate services would be the way to go. Easier to test since the service is only doing one thing, and more decoupled.

bzwei: Yes in many cases one service is doing one thing, but why do we have to limit to one method? For example `create` and `create!` are very closely related, it makes sense to group them together.
For a real example, our `JbpmProcessService` has `#start` and `#signal` methods. They are very similar and even share some code.

About appending service to the class name - yes, all classes are placed under service folder which makes the appendix seem redundant. But in the code where the service class is called we can't easily tell the class is from the service folder. For readability purpose I think the appendix helps.

eclarizio: @bzwei I mostly agree with what you're saying, which is why I wasn't really a huge fan of the "process" call in the first place, was just laying out some of the reasons I think that we were using that method. I think your `JbpmProcessService` may be a poor example, because sure, the methods *look* the same, but `#start` has *way* more complexity, due to using the private method `#process_options` which in turn uses `#enhance_groups`, etc. However I understand your point. I still think it would be cleaner if that logic was separated out, it feels like one class is doing too much at that point.

The only place I'd expect service classes themselves to be called is either in the controller, or, if they are more of an infrastructure service, in a model, and I don't know if it's that important that we know by the name that it is a service. I would just assume anything being called by a controller or a model is some kind of service, regardless of if it has the "Service" suffix. Again, I don't really mind that we tack on `Service` to the end, just sharing why I think we currently don't.

app/services/workflow_create_service.rb
bzwei: The above comment applies to update and destroy operations. For creation there is no reload. But retry helps to find newest available sequence thus avoid duplicates.

bzwei:
@gmcculloug @eclarizio Can we move on reviewing the logic or merging this PR since it is an important fix? We can keep the service class coding style as is for now to sync with other existing service classes. A follow-up PR can change the style for all classes if we decide that is the right way.

PR #369, opened 2020-06-03
Separate errors of external services into more detailed categories

hsong-rh:
https://projects.engineering.redhat.com/browse/SSP-1301

In this PR:
1. Separated `NetworkError` and `TimedOutError` from common `KieError` for PAM service;
2. Separated `NetworkError`, `TimedOutError` and `NotAuthorizedError` from `Faraday` exceptions, marked other errors as `TaggingError`, for remote tagging services

### DIFF
--- a/app/services/remote_tagging_service.rb
+++ b/app/services/remote_tagging_service.rb
@@ -6,6 +6,7 @@ class RemoteTaggingService
   # TODO: Support proper pagination of tags from Faraday since
   # we are not using the generated client here.
   QUERY_LIMIT = 1000
+
   def initialize(options)
     @app_name = options[:app_name]
     @object_type = options[:object_type]
@@ -46,7 +47,7 @@ class RemoteTaggingService
 
   def service_url
     match = self.class.remotes.detect { |item| item[:app_name] == @app_name && item[:object_type] == @object_type }
-    raise "No url found for app #{@app_name} object #{@object_type}" unless match
+    raise Exceptions::InvalidURLError.new("No url found for app #{@app_name} object #{@object_type}") unless match
 
     match[:url].call
   end
@@ -75,7 +76,7 @@ class RemoteTaggingService
     if response.status == 403
       raise Exceptions::NotAuthorizedError, response.reason_phrase
     else
-      raise "#{message_prefix} #{response.reason_phrase}" unless VALID_HTTP_CODES.include?(response.status)
+      raise Exceptions::TaggingError.new("#{message_prefix} #{response.reason_phrase}") unless VALID_HTTP_CODES.include?(response.status)
     end
   end
 
--- a/app/services/remote_tagging_service.rb
+++ b/app/services/remote_tagging_service.rb
@@ -2,7 +2,6 @@ require 'faraday'
 require_relative 'mixins/tag_mixin'
 class RemoteTaggingService
   include TagMixin
-  VALID_HTTP_CODES = [200, 201, 202, 204, 304].freeze
   # TODO: Support proper pagination of tags from Faraday since
   # we are not using the generated client here.
   QUERY_LIMIT = 1000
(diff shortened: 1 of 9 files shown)


app/services/remote_tagging_service.rb
```diff
@@ -6,6 +6,10 @@ class RemoteTaggingService
   # TODO: Support proper pagination of tags from Faraday since
   # we are not using the generated client here.
   QUERY_LIMIT = 1000
+
+  class NoURLError   < StandardError; end
```
bzwei, line 10: All Errors are defined in `lib/exceptions.rb`

lib/kie/service.rb
```diff
@@ -1,11 +1,17 @@
 require 'kie_client'
 module Kie
+  class NetworkError  < StandardError; end
```
bzwei, line 3: Move into `lib/exceptions.rb`. The error can be used by all external calls resulting network error or timeout.

app/services/remote_tagging_service.rb
```diff
@@ -75,7 +79,7 @@ def check_for_exceptions(response, message_prefix)
     if response.status == 403
       raise Exceptions::NotAuthorizedError, response.reason_phrase
     else
-      raise "#{message_prefix} #{response.reason_phrase}" unless VALID_HTTP_CODES.include?(response.status)
+      raise TaggingError.new("#{message_prefix} #{response.reason_phrase}") unless VALID_HTTP_CODES.include?(response.status)
```
bzwei, line 82: Can we detect connection and timeout errors like how we make Kie api calls? If yes, we can mark them service_unavailable instead of bad request.

hsong-rh: I went through the underlying `faraday` gem used by remote tagging, it's hard to differentiate the `connection` or `timeout` errors based on the response status. So the answer is no

app/services/remote_tagging_service.rb
```diff
@@ -46,7 +47,7 @@ def object_url
 
   def service_url
     match = self.class.remotes.detect { |item| item[:app_name] == @app_name && item[:object_type] == @object_type }
-    raise "No url found for app #{@app_name} object #{@object_type}" unless match
+    raise Exceptions::InvalidURLError.new("No url found for app #{@app_name} object #{@object_type}") unless match
```
bzwei, line 50: InvalidURLError is misleading to me. `Exceptions::UserError` is more appropriate because we cannot find a matched preconfigured URL based on user provided `app_name` and `object_type`. To users they do not know there is an internal URL involved.

lib/kie/service.rb
```diff
@@ -6,6 +6,9 @@ def self.call(klass, options)
       yield init(klass)
     rescue KieClient::ApiError => err
       Rails.logger.error("KieClient::ApiError #{err.message} ")
+
+      raise Exceptions::TimedOutError.new('Connection timed out') if err.code.nil?
+      raise Exceptions::NetworkError.new(err.message) if err.code.zero?
```
eclarizio, line 11: Do we expect any other services that we connect to (such as RBAC?) to ever potentially have these same errors? I know we don't currently distinguish, but these seem generic when they are really meant for the Kie service, right? Maybe itt makes more sense to have `Exceptions::KieError::TimedOut` and `Exceptions::KieError::Network` as the actual errors?

hsong-rh: @eclarizio Similar logic for RBAC already handled in api-common. `TimedOutError` and `NetworkError` are changed from `Kie::TimedOutError` and `Kie::NetworkError`. The reason is that they can be used by all external services.

bzwei: All the errors come from HTTP calls. They can be commonly used by all eternal http based api communication.

eclarizio: @hsong-rh Ah ok, so we do expect to use these in other places then? Sounds good 👍

app/services/remote_tagging_service.rb
hsong-rh: Changed to catch Faraday exceptions directly

PR #370, opened 2020-06-04
Bump puma to v4.3

gmcculloug:
Noticed we were lagging behind the other services with the puma version. This version also addresses recent vulnerabilities.

Also dropped version from test gems factory_bot_rails and rspec-rails.

Matching
https://github.com/RedHatInsights/topological_inventory-api/blob/master/Gemfile#L16
https://github.com/RedHatInsights/catalog-api/pull/758

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -11,7 +11,7 @@ gem 'manageiq-loggers',    '~> 0.2'
 gem 'manageiq-messaging',  '~> 0.1'
 gem 'pg',                  '~> 1.0', :require => false
 gem 'prometheus-client',   '~> 0.8.0'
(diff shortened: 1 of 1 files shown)


bzwei:
@gmcculloug Not sure why we still have security warnings after we upgraded the version.

gmcculloug:
@bzwei Might be a Hakiri issue, but I'm switching the version to just be '>= 4.3.5' which we are doing elsewhere: https://github.com/RedHatInsights/topological_inventory-api/blob/master/Gemfile#L16

gmcculloug:
Not sure why travis does not want to pick up the tests.  🤷‍♂️

PR #372, opened 2020-06-04
Fix mis-spelled error class name

bzwei:
We renamed an error class name in #367 that needs to be changed in the error conversion list.


bzwei:
Add a spec test to make sure all class names in string is are valid classes and symbols are valid http status.

Don't know why travis did not run.

PR #374, opened 2020-06-15
Rescue Active::DeadLocked error and retry

bzwei:
Also changes to order-by-id to reduce the possibility of deadlocking.

The problem was reported in QA env, not reproducible by any spec unit tests. Add rescue for deadlocked error in all workflow services that may modify sequence.

https://projects.engineering.redhat.com/browse/SSP-1592

### DIFF
--- a/app/models/workflow.rb
+++ b/app/models/workflow.rb
@@ -71,13 +71,13 @@ class Workflow < ApplicationRecord
   end
 
   def change_sequences_to_negative(startp, endp, delta)
-    query = self.class.where(table[:sequence].gteq(startp))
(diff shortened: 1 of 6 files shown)


bzwei:
@eclarizio deadlock can be resolved by retry.

slemrmartin:
@bzwei is the reordering in bulk updates really helping with deadlocks? We have deadlock problem in topological-inventory's core and I want to fix it this way too

bzwei:
@slemrmartin To avoid deadlock you have to ensure multiple records always update in the same order. Also retry may help.

PR #375, opened 2020-06-16
Add object_dependencies into workflow metadata

hsong-rh:
This PR adds another attribute `object_dependencies` into the metadata of `workflow`. It stores the dependent categories, which is used by UI to post warning message before workflow is deleted.

New format of metadata looks like: 
```yaml
{
  "user_capabilities"=>{
    "create"=>true,
    "unlink"=>true,
    "show"=>true,
    "update"=>true,
    "destroy"=>true,
    "link"=>true
  },
  "object_dependencies"=>{
    "catalog"=>[
      "Portfolio",
      "PortfolioItem"
    ],
    "sources"=>[
      "Source"
    ]
  }
}
```
where `catalog`, `sources` are the `app` names, the array value contains the dependent categories.

### DIFF
--- a/app/models/workflow.rb
+++ b/app/models/workflow.rb
@@ -9,7 +9,7 @@ class Workflow < ApplicationRecord
   has_many :tag_links, :dependent => :destroy, :inverse_of => :workflow
 
   validates :name, :presence => true, :uniqueness => {:scope => :tenant}
-  validates :sequence, :uniqueness => { scope: :tenant_id }
+  validates :sequence, :uniqueness => {:scope => :tenant_id}
 
   before_validation :new_sequence, :on => :create
   before_validation :adjust_sequences, :on => :update
@@ -25,7 +25,20 @@ class Workflow < ApplicationRecord
     template&.signal_setting.present?
   end
 
+  def metadata
+    super.merge(:object_dependencies => object_dependencies)
+  end
+
   private
+
+  def object_dependencies
+    {}.tap do |dependencies|
+      tag_links.pluck(:app_name, :object_type).uniq.each do |key, value|
+        dependencies[key] ||= []
+        dependencies[key] << value
+      end
+    end
+  end
 
   def table
     self.class.arel_table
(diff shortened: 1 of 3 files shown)


app/models/workflow.rb
```diff
@@ -25,8 +25,18 @@ def external_signal?
...
 
+  def metadata
+    super.merge(:object_dependencies => object_dependencies)
+  end
+
   private
 
+  def object_dependencies
+    Hash.new { |h, k| h[k] = [] }.tap do |h|
+      tag_links.where(:workflow_id => id).pluck(:app_name, :object_type).uniq.each { |k, v| h[k] << v }
```
gmcculloug, line 36: @bzwei @hsong-rh Are there performance concerns with this running as part of an API  call returning a list of 50 or 100 workflows?  (UI caps returns at 100 right now.)

@eclarizio Thoughts on this?  I imagine we could look to use the ancillary_metadata approach here, but not sure we need to go that path just yet.

hsong-rh: @gmcculloug I removed the unnecessary `where` query. it does have impact on performance. But comparing with the other approach of adding separate endpoint to query these information, which means extra API calls, I feel this impact is acceptable.

gmcculloug: I would agree with that.  The table looks like it is indexed properly and we would not expect a lot of data here so probably would not see a that big of an impact overall.

Still curious if others have thoughts on this approach.  @eclarizio @mkanoor

app/models/workflow.rb
```diff
@@ -25,8 +25,18 @@ def external_signal?
...
+  def metadata
+    super.merge(:object_dependencies => object_dependencies)
+  end
+
   private
 
+  def object_dependencies
+    Hash.new { |h, k| h[k] = [] }.tap do |h|
+      tag_links.pluck(:app_name, :object_type).uniq.each { |k, v| h[k] << v }
+    end
```
eclarizio, line 37: This is kind of difficult to digest. What do we think of this? Better? Worse?
```
Hash.new.tap do |dependencies|
  tag_links.pluck(:app_name, :object_type).uniq.each do |key, value|
    dependencies[key] ||= []
    dependencies[key] << value
  end
end
```

hsong-rh: @eclarizio 👍

bzwei:
LGTM. @hsong-rh you need to update openapi_spec.json too.

PR #377, opened 2020-06-23
Fall back to use default version for openapi-generator-cli validation

bzwei:
Sometimes script openapi-generator-cli.sh fails to fetch openapi-generator's latest version. We then fall back to use current latest 4.3.1.

### DIFF
--- a/.travis_scripts/openapi-validator.sh
+++ b/.travis_scripts/openapi-validator.sh
@@ -2,13 +2,12 @@
 set -eu
 set -o pipefail
 
-if ! ./openapi-generator-cli version
-then
-  echo "openapi-generator-cli version failed, likely caused by a mvn build failure. Retry once."
-  ./openapi-generator-cli version
+if ! ver=$(./openapi-generator-cli version | tail -1); then
+  ver="4.3.1"  
+  echo "Use default version ${ver} to validate"
 fi
 
 for entry in ./public/doc/openapi-3-v*.json
 do
-  ./openapi-generator-cli validate -i "$entry"
+  OPENAPI_GENERATOR_VERSION=${ver} ./openapi-generator-cli validate -i "$entry"
 done


.travis_scripts/openapi-validator.sh
```diff
@@ -2,13 +2,23 @@
 set -eu
 set -o pipefail
 
-if ! ./openapi-generator-cli version
+ver=4.3.1
+if ./openapi-generator-cli version
```
mkanoor, line 6: @bzwei This line is going to do the download.
Why are we not capturing the output of the command 
```
VERSION=`./openapi-generator-cli version`
```
And using that along with
```
OPENAPI_GENERATOR_VERSION=${VERSION} ./openapi-generator-cli validate -i "$entry"
```
I had mentioned this here https://github.com/RedHatInsights/approval-api/pull/358#discussion_r427584395

bzwei: @mkanoor I was thinking parsing from the jar filename to get version is more accurate in case the version command may produce another format other than pure version numbers in the future. 
Anyway I now switch to directly capture output.

mkanoor:
Something like this works on RHEL
```
#! /bin/sh
VERSION=`/tmp/installer_scripts/openapi-generator-cli.sh version|tail -1`
echo $?
echo $VERSION
```
You would have to check if the command exited with 0 before you use the $VERSION. Also the path to the script is fully qualified in my case the script was in the directory /tmp/installer_scripts

.travis_scripts/openapi-validator.sh
```diff
@@ -2,13 +2,12 @@
 set -eu
 set -o pipefail
 
-if ! ./openapi-generator-cli version
-then
-  echo "openapi-generator-cli version failed, likely caused by a mvn build failure. Retry once."
-  ./openapi-generator-cli version
+if ! ver=$(./openapi-generator-cli version | tail -1); then
```
mkanoor, line 5: @bzwei We should be handling the failure case. If the jar cannot be downloaded the version call is going to fail and we have to reattempt to run the version command again. If it fails after 2 or 3 attempts we should fail the script and not  go on to the validation phase.

mkanoor: @bzwei Another idea is can we post this shell script which does the validation in its own repo so we make the changes in one place. And from other travis.yml file for catalog, approval, topology, sources we can just fetch the script from this new repo. So that way we dont have to make these changes in every repo. In every repo the .travis.yml will fetch the file and execute it.

bzwei: The failed the case is already taken care of. Here we will fall back to default version.
In the past we discovered a retry does not help. The script does not download the jar, instead it builds the jar with a given version. It sometimes fails to decide the latest version because it needs to download  https://api.github.com/repos/openapitools/openapi-generator/releases to parse the latest version. This file sometime is not available for download and a retry in short time does not help. Therefore I decide to fallback to a default version.

mkanoor: @bzwei This is the line that downloads the jar on a brand new build env.
https://github.com/OpenAPITools/openapi-generator/blob/6b877efe1f372163ecc3739f175d9e6f911344e4/bin/utils/openapi-generator-cli.sh#L58

There is no cache on that machine, if that download fails we need something to handle that.
Just setting the version to 4.3.1 doesn't help. When the next call is made it will get to the same code again and try to fetch the file again.

If the jar cannot be downloaded for some reason we need to guard against that.

bzwei: @mkanoor All our failures found so far rooted at https://github.com/OpenAPITools/openapi-generator/blob/6b877efe1f372163ecc3739f175d9e6f911344e4/bin/utils/openapi-generator-cli.sh#L31. This file determines the latest version to use. 

https://github.com/OpenAPITools/openapi-generator/blob/6b877efe1f372163ecc3739f175d9e6f911344e4/bin/utils/openapi-generator-cli.sh#L58 is a maven command to build (download) all dependencies, but it is a one time thing unless the desired version changes. So far all the failures in this step were caused by an empty version. 

But every run always downloads the release file in L31. This download can fail. I opened an issue in https://github.com/OpenAPITools/openapi-generator/issues/6748. I even saw one time the first run succeeded but the second run failed because the second run failed to download the release file. 

The solution here for us is to make sure we don't end with an empty version, By using either latest or a default version L58 will not likely fail.

gmcculloug: @bzwei I'm ok with this.  What is the impact if we get behind in versions?

Is it only an issue if there are bugs in the validator that are fixed?  Looking at the [openapi-generator tags](https://github.com/OpenAPITools/openapi-generator/tags) listing I can see us getting behind in patch releases but I'm not sure there is much of a penalty when we do.

bzwei: We can try this for a while to see how it is doing. First falling back a default version is a back up plan. In most of time we are still using latest version. Second I would expect old version's validation function to work well until we start to use some new features in openapi spec.

gmcculloug:
I opened an issue for that work: https://github.com/RedHatInsights/insights-api-common-rails/issues/201

PR #379, opened 2020-07-06
Handle a special case when moving last sequence higher

bzwei:
This is a comer case. Do not adjust neighboring sequences when attempt to move the last sequence higher.

https://projects.engineering.redhat.com/browse/SSP-1650

### DIFF
--- a/app/models/workflow.rb
+++ b/app/models/workflow.rb
@@ -80,7 +80,7 @@ class Workflow < ApplicationRecord
 
     if sequence > sequence_was
       sequence_lower(sequence_was, sequence)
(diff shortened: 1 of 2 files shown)


app/models/workflow.rb
```diff
@@ -80,7 +80,7 @@ def adjust_sequences
```
hsong-rh, line 80: How about we `return` from here directly for this case? I feel it's more clear than current fix

bzwei: it is a temporary fix. The whole logic is going away soon.

PR #381, opened 2020-07-08
Send workflow deleted event in service class

bzwei:
Fix the bug that we send workflow_deleted event after every commit since it is only needed for destroy operation.

It is more appropriate to send the event in the service class rather than in the model class

### DIFF
--- a/app/services/workflow_delete_service.rb
+++ b/app/services/workflow_delete_service.rb
@@ -9,6 +9,7 @@ class WorkflowDeleteService
     begin
       retries ||= 0
       Workflow.find(workflow_id).destroy!
+      EventService.new(nil).workflow_deleted(workflow_id)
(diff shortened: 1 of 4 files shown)


bzwei:
@hsong-rh In the service class we send event after `destroy!` which guarantees the commit has succeeded.

app/services/workflow_delete_service.rb
```diff
@@ -9,6 +9,7 @@ def destroy
     begin
       retries ||= 0
       Workflow.find(workflow_id).destroy!
+      EventService.new(nil).workflow_deleted(workflow_id)
```
eclarizio, line 12: I know you just moved this, so I'm not going to hold up the PR for it, but it feels weird that we're passing in `nil` here. I think we should just use `initialize(request = nil)` in the definition so that we would do `EventService.new.workflow_deleted` here. That can be a separate small refactor change, though.

PR #385, opened 2020-07-22
EventService takes default value in initilize method

bzwei:
This is a follow-up PR to address comments in #381


bzwei:
@eclarizio @hsong-rh Added a commit to fix a bug about how to set headers when request is not present.

PR #406, opened 2020-09-12
Fails request if RBAC error raises during runtime

hsong-rh:
RBAC error during runtime will cause approval request pending forever. This PR rescues such errors and updates request with failed state.

https://projects.engineering.redhat.com/browse/SSP-1814

### DIFF
--- a/app/services/mixins/group_validate_mixin.rb
+++ b/app/services/mixins/group_validate_mixin.rb
@@ -22,8 +22,11 @@ module GroupValidateMixin
 
     begin
       group = ensure_group(request.group_ref, request.context)
-    rescue Exceptions::UserError, RBACApiClient::ApiError, StandardError => e
+    rescue Exceptions::UserError => e
       error_action(request, e.message)
+    rescue => e
+      error_action(request, "Internal error while validating a group. Check log for details.")
+      Rails.logger.error(["Caught an error while validating a group before forwarding the request to PAM.", e.message, *e.backtrace].join($RS))
     end
 
     return false unless group
@@ -71,7 +74,7 @@ module GroupValidateMixin
              end
            rescue Exceptions::UserError
              "#{old_name}(Group does not exist)"
-           rescue StandardError
+           rescue
              old_name
            end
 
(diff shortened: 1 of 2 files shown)


app/services/mixins/group_validate_mixin.rb
```diff
@@ -24,6 +24,12 @@ def runtime_validate_group(request)
       group = ensure_group(request.group_ref, request.context)
     rescue Exceptions::UserError => e
       error_action(request, e.message)
+    rescue RBACApiClient::ApiError => e
+      error_action(request, e.message)
```
bzwei, line 28: Do not send `e.message` to action because it is meant to be shown to user. 
Consider the following for all Non UserError
```
  error_action(request, "Internal error while validating a group. Check log for details.")
  Rails.logger.error(["Caught an error while validating a group before forwarding the request to PAM.", e.message, *e.backtrace].join($RS))
```

eclarizio: I think you meant `it is meant to be shown to user` to be `it is *not* meant to be shown to user`, and I agree 👍

hsong-rh: I got your means 😄

eclarizio (approved):

app/services/mixins/group_validate_mixin.rb
```diff
@@ -71,7 +74,7 @@ def validate_approver_group_no_raise(uuid, old_name)
              end
            rescue Exceptions::UserError
              "#{old_name}(Group does not exist)"
-           rescue StandardError
+           rescue
```
bzwei, line 77: rubocop used to raise a flag if rescue any error.

app/services/mixins/group_validate_mixin.rb
```diff
@@ -24,6 +24,9 @@ def runtime_validate_group(request)
       group = ensure_group(request.group_ref, request.context)
     rescue Exceptions::UserError => e
       error_action(request, e.message)
+    rescue => e
```
bzwei, line 27: rubocop used to raise a flag if rescue any error.

bzwei (approved):

PR #409, opened 2020-09-22
Remove sequence from workflows

bzwei:
Sequence has been replaced by internal_sequence, and UI has switched to use internal_sequence. Now it is time to officially remove sequence from workflows.

The original thought was to introduce a new API version. Our current API version is v0.1.2. Since we have made database change, it is impossible to support both v0.1.2 and v0.1.3. And if we simply replace v0.1.2 by v0.1.3, it is impossible to make all other services such as catalog-api and UI to update their client version simultaneously. The best option is to keep the existing version.

https://issues.redhat.com/browse/SSP-1843


eclarizio (approved):
Looks fine to me. We could fix those rubocop suggestions, I guess, but in a migration I don't really mind.

hsong-rh (approved):

PR #410, opened 2020-09-29
Fix reposition api return code and required params

bzwei:
https://issues.redhat.com/browse/SSP-1856
https://issues.redhat.com/browse/SSP-1857

Fix openapi spec for workflow reposition/link/unlink. All of them should return 204.

Raise 400 error if no param is given in the body of reposition api.


hsong-rh (approved):
LGTM 👍

eclarizio (approved):

PR #413, opened 2020-10-01
Upgrading the insights-api-common gem to 4.1.4 version.

pkomanek:
Upgrading the insights-api-common gem to 4.1.4 version that no longer allows redirects from major versions on a POST.

This PR is based on https://issues.redhat.com/browse/RHCLOUD-9585.


bzwei:
@pkomanek This change is currently not needed.

PR #414, opened 2020-10-27
Update codeclimate and rubocop yml to use manageiq-style

eclarizio:
Just mimicking what we did in catalog-api so that any future PRs for approval-api get updated with the "right" style guide. I imagine this may eventually change to use the RedHatInsights one that Jacob is putting together: https://github.com/RedHatInsights/insights-api-common-rails/pull/210

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -2,9 +2,9 @@
 version: '2'
 prepare:
   fetch:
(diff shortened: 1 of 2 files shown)


eclarizio:
@bzwei This one also has the weird issue of the codeclimate build linking to the wrong PR. https://codeclimate.com/github/RedHatInsights/approval-api/builds/577 is the correct build for this PR, and it passed. So very strange 🤷

hsong-rh (approved):

PR #420, opened 2021-01-28
Not to delete so files

bzwei:
The old Dockerfile deletes .so files. Bundle install fetches and installs latest nokogiri 1.11.1. When the pod starts it fails immediately with `LoadError: cannot load such file -- nokogiri/nokogiri"`. 

I have tested that the pod starts properly if we leave those files un-removed.

@lindgrenj6 Not sure why the Dockerfile worked until now. I feel it is safe to leave the library files un-removed. Please review.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,7 +18,7 @@ COPY Gemfile $WORKDIR
 RUN echo "gem: --no-document" > ~/.gemrc && \
     gem install bundler --conservative --without development:test && \
     bundle install --jobs 8 --retry 3 && \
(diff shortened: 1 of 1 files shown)


hsong-rh (approved):

lindgrenj6:
Huh yeah I didn't know approval had this. 

I thought I removed it from all of the containers when we switched to rdkafka, that's when I noticed it last. 

Either way this is a safe change. It will just increase the image size by 10-20MB (nothing when the image is 800MB!)

PR #424, opened 2021-02-05
Add log message in remote tagging

hsong-rh:


hsong-rh:
No need to add, closed

PR #425, opened 2021-02-23
Do not raise error if signal fails

bzwei:
This addresses two issues
1. Upon each request finishing we need to send a signal to PAM telling it the process finished. If the signal fails it should not affect our approval request state. We can thus safely ignore the error instead of turning the request's state to failed.
2. `RequestUpdateService#finish_request` skips sending the signal if `request_finished` is true. However this method in the context always returns true so that signal is never triggered. Removing this check will force the signal to be sent. Even if it fails it has no impact to the request itself due to the change in 1.

### DIFF
--- a/app/services/jbpm_process_service.rb
+++ b/app/services/jbpm_process_service.rb
@@ -21,8 +21,8 @@ class JbpmProcessService
       bpm.signal_process_instance(options['container_id'], request.process_ref, options['signal_name'], :body => signal_options(decision))
     end
   rescue => err
-    ActionCreateService.new(request.id).create(:operation => Action::ERROR_OPERATION, :processed_by => 'system', :comments => err.message)
-    raise
+    # Signaling PAM is the last step after the approval has finished. Ignore the error if signal fails
+    Rails.logger.error("Failed to signal PAM with error: #{err.message}")
   end
 
   private
--- a/app/services/request_update_service.rb
+++ b/app/services/request_update_service.rb
@@ -150,7 +150,6 @@ class RequestUpdateService
 
   # complete the external approval process if configured
   def finish_request(decision)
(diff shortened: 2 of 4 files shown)


app/services/jbpm_process_service.rb
```diff
@@ -21,8 +21,8 @@ def signal(decision)
       bpm.signal_process_instance(options['container_id'], request.process_ref, options['signal_name'], :body => signal_options(decision))
     end
   rescue => err
-    ActionCreateService.new(request.id).create(:operation => Action::ERROR_OPERATION, :processed_by => 'system', :comments => err.message)
-    raise
+    # Signaling PAM is the last step after the approval has finished. Ignore the error if signal fails
+    Rails.logger.error("Failed to signal PAM with error: #{err.message}")
```
eclarizio, line 25: Since I'm not as well versed in approval, this might be a stupid question, but basically is your comment here implying that is why line 24 is no longer needed?

app/services/request_update_service.rb
```diff
@@ -150,7 +150,6 @@ def notify_request
 
   # complete the external approval process if configured
   def finish_request(decision)
-    return if request_completed?(decision)
```
eclarizio, line 153: Is there no spec covering the removal of this line or is it being somehow tested within the `jbpm_process_service_spec` that you changed?

app/services/jbpm_process_service.rb
bzwei: yes

app/services/request_update_service.rb
bzwei: I added a spec in request_update_service_spec.rb

eclarizio (approved):

PR #426, opened 2021-02-24
Send group_finished event after update

bzwei:
https://issues.redhat.com/browse/SSP-1779

The problem with group_finished event was the decision is always `undecided`. The event was sent before the request was saved. Now move the sending after the request has been updated.

Also fixed the issue that request_started event were sent multiple times when multiple groups were involved.

### DIFF
--- a/app/services/request_update_service.rb
+++ b/app/services/request_update_service.rb
@@ -26,7 +26,7 @@ class RequestUpdateService
 
     EventService.new(request).request_started if request.root?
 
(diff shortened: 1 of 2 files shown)


eclarizio (approved):
Logic makes sense to me, but @hsong-rh please review before I merge since you have more involved knowledge here to make sure I'm not missing anything.

hsong-rh (approved):

PR #447, opened 2021-10-04
Use dnf command to do cleaning

bzwei:
Align with https://github.com/RedHatInsights/catalog-api/blob/main/Dockerfile#L15
Hopefully it can reduce the number of dependent packages in the image.


eclarizio:
@bzwei Is the ci.ext.devshift.net PR build failure check a concern?

eclarizio (approved):

bzwei:
> @bzwei Is the ci.ext.devshift.net PR build failure check a concern?

This build never works for us. You can ignore it.

